PR #4, opened 2020-03-11
Fix test suite

damiendoligez:


lemmy:
@damiendoligez Never mind the "All checks have failed".  It was a bug in the Github action that has been fixed with 1baca83.

lemmy:
@damiendoligez 255467d breaks the Github action ([log](https://pipelines.actions.githubusercontent.com/epbWR4RpwQVwMzaJNlYI312WXPUza4f4rW2oifqVVNih7inrQA/_apis/pipelines/1/runs/6/signedlogcontent/4?urlExpires=2020-03-11T23%3A16%3A44.4302541Z&urlSigningMethod=HMACV1&urlSignature=UDIwukdXoSCZkc4Ych2Wqjtv6aZ9ELECKjFRqqQ4fQQ%3D)).  With 255467d reverted, there remain 9 test failures in `testall` ([log](https://pipelines.actions.githubusercontent.com/epbWR4RpwQVwMzaJNlYI312WXPUza4f4rW2oifqVVNih7inrQA/_apis/pipelines/1/runs/7/signedlogcontent/3?urlExpires=2020-03-11T23%3A16%3A15.3771386Z&urlSigningMethod=HMACV1&urlSignature=9FLkxA1USZvvZH5wa5Ifgb0Ky%2BHCOIC0J7DA%2BLIcofg%3D)).

test/TOOLS/do_tests
```diff
@@ -10,7 +10,8 @@ if [ "$1" = "smt_tests" ]
   then export ARGS="$ARGS --method smt"
 fi
 
-export TLAPM="$BASEDIR/../tlapm -I $ARGS"
+export TLAPM="$BASEDIR/bin/tlapm"
```
quicquid, line 13: This leads to tlapm being called from the test/bin directory. Even when I put the tlapm binary there, the missing -I flag fails all tests that extend TLAPS. Is this the way we want it?

test/regression_tests/consensus/consensus_test.tla
```diff
@@ -22,7 +22,12 @@ THEOREM Invariance == Spec => []Inv
 <1>1 Init => Inv
   BY CardinalityZero, SMT DEFS Init, Inv
 <1>2. Inv /\ [Next]_chosen => Inv'
-  BY CardinalityOne, SMT DEFS Next, Inv
+  \* BY CardinalityOne, SMT DEFS Next, Inv
+  <2>1. Inv /\ UNCHANGED chosen => Inv'
+    BY SMT DEF Inv
+  <2>2. Inv /\ Next => Inv'
+    BY CardinalityOne, SMT DEF Next, Inv
```
quicquid, line 29: This step doesn't go through on my machine, I had to expand this to:

    <2>2. Inv /\ Next => Inv'
      <3>1. Inv /\ Next => (chosen \subseteq Value)' BY CardinalityOne, SMT DEF Next, Inv 
      <3>2. Inv /\ Next => IsFiniteSet(chosen)' BY CardinalityOne, SMT DEF Next, Inv
      <3>3. Inv /\ Next => (Cardinality(chosen) \leq 1)' BY CardinalityOne, SMT DEF Next, Inv
      <3> QED BY <3>1, <3>2, <3>3 DEF Inv

The conjunction of <3>2 and <3>3 doesn't go through here either.

muenchnerkindl: The proof as written by Damien works for me – but I suggest replacing the `Sets' module by the library module `FiniteSetTheorems' and CardinalityZero / CardinalityOne by the corresponding library lemmas.
[consensus_test.txt](https://github.com/tlaplus/tlapm/files/4550306/consensus_test.txt)

test/TOOLS/do_tests
damiendoligez: In fact it should be written somewhere, but the correct setup is to use the installer to install the whole system into the test directory:
```
tlaps-1.4.5-*-inst -d /home/.../test
```
This means that we test not only the program itself, but also the installation process.

test/regression_tests/consensus/consensus_test.tla
damiendoligez: @muenchnerkindl I have implemented your suggestion and then simplified the proof further. It goes through quite quickly on my machine.

PR #18, opened 2020-12-03
update version number and remove old svn-based commit numbers

damiendoligez:
Release 1.5.0 will be the first with `ENABLED` and `\cdot` handling.


lemmy:
FYI: [main.yml and pr.yml](https://github.com/tlaplus/tlapm/tree/master/.github/workflows) both have to be updated to reflect the new version number.

PR #19, opened 2020-12-03
classify tests into fast/medium/slow

damiendoligez:


johnyf:
These changes look good to me. A relevant change could be to add a rule `medium` to the file `test/Makefile` for running tests under the directory `test/medium`, similarly to the rules `fast` and `slow`.

Also, there are two files under the directory `test/soundness_tests` that are not part of the tests, and could possibly be turned into fast tests that fail (the files are not configured to check for failure, I assume they are supposed to not be provable based on reading their content).

johnyf:
Thank you for adding the rule `medium` to the `Makefile`.

PR #22, opened 2020-12-08
BUG: lex keywords in WF_, SF_ subscripts

johnyf:
This change ensures that `TRUE` and `FALSE`
can be used as fairness subscripts, for
example `WF_TRUE(TRUE)`.

Previously, the string `WF_TRUE` was lexed
as punctutation `WF_` followed by an
identifier `TRUE`, instead of the keyword `TRUE`.

This commit adds also a test module for this case.

### DIFF
--- a/src/alexer.mll
+++ b/src/alexer.mll
@@ -39,6 +39,16 @@ let namechar = (letter | numeral | '_')
 
 let name     = namechar* letter namechar*
 
+let keyword = (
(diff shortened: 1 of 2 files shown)


damiendoligez (approved):
This is a simple bug fix that looks obviously correct. I'll merge immediately.

johnyf:
Thank you for merging these changes.

PR #24, opened 2020-12-10
`INSTANCE` within `LET`, `ENABLED` in parametrized `INSTANCE`, `ENABLED` in instantiated operators with parameters, and related changes

johnyf:
These changes include bug fixes:

- cdf6240de4ddd2466a995886950c260477ecf3a7: recomputing the levels of operators after instantiation, because parametric `INSTANCE` changes the arity of operators, and also can change the level of declared identifiers. A change in arity makes the level information of operators outdated, and requires recomputing the operator's level information.

- e27d639689a211d47e3986470c570497b12eb665: bind primed occurrences of variables before instantiation, when these appear in scope of `ENABLED` inside the definition of an operator that takes arguments. This change:
  - ensures correct binding of primed occurrences of variables when instantiating, for example, the operator `A(z) == ENABLED (x' # y')` (as in the added test module)
  - raises an error during operator instantiation if an operator parameter occurs inside the scope of `ENABLED`, for example:
  ```tla
  ---- MODULE RaisesError ----
  ---- MODULE Inner ----
  A(x) == ENABLED (x')
  ======================
  M == INSTANCE Inner
  ============================

- 2ec2b79911436166d994e1ee5d01b7a3ed259b35: in the test scripts, tracking nesting of submodules inside a module, to decide which horizontal line `=====*` terminates the file's module, so that the test commands be read after that `====*` line

- d0b800e58cdb758b7edef04345719dd00f868182: show `VARIABLE` for single variable, `VARIABLES` for multiple variables when formatting modules

- dbe57c610278d85a24abeb079c986b165cecf28f: load modules that occur inside an `EXTENDS` that itself appears inside a submodule

- d114852d3dfcaff000d5c9096ac90e5fb44030e4: module units need to be correctly shifted when instantiating. The shift depends on how many hypotheses each of the omitted module units introduced in the module context (and this number is not `1`, which was previously used)

- a5035a822f6552a956b8d78b82792f30ea65ceff, 4815b340f10f5c650ece9826c5b84515c551ecc7: allow `INSTANCE` statements to occur within `LET`: there were errors when the instantiated module (e.g., `TLAPS`) included backend pragmas, because only definitions were expected in normalization of `LET` expressions, and in fingerprinting of `LET` expressions.

Test cases are also added for each change. Please find more details in the commits messages.

### DIFF
--- a/src/backend/fingerprints.ml
+++ b/src/backend/fingerprints.ml
@@ -516,8 +516,10 @@ and fp_let counthyp countvar stack buf ds e =
             Stack.push stack (Identvar hint.core, ref false);
             spin stack ds;
             ignore(Stack.pop stack)
+        | Instance _ ->
+            Errors.bug ~at:d "Backend.Fingerprints.fp_let: INSTANCE 3"
         | _ ->
-            Errors.bug ~at:d "Backend.Fingerprint.fp_let: INSTANCE 3"
+            Errors.bug ~at:d "Backend.Fingerprints.fp_let: other"
       end
   in
   spin stack ds
--- a/src/backend/fingerprints.ml
+++ b/src/backend/fingerprints.ml
@@ -510,7 +510,8 @@ and fp_let counthyp countvar stack buf ds e =
   let rec spin stack = function
     | [] -> fp_expr counthyp countvar stack buf e
     | d :: ds -> begin match d.core with
-        | Operator (hint, e) ->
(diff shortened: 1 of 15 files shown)


damiendoligez (approved):
I've reviewed these changes and they all look good to me.

johnyf:
Thank you for reviewing the changes and approving this pull request, the changes are now merged.

PR #49, opened 2021-09-28
ENH: convert syntax tree to JSON

johnyf:
These changes serialize the abstract syntax tree as JSON.


ahelwer:
@johnyf do you think something like this would be better implemented by sticking PPX macros onto our data structures?

ahelwer:
This seems like useful functionality I'd like to play around with, would you be able to rebase and fix the conflicts?

PR #55, opened 2021-11-23
Smt fix

adef-inr:
Contains two soundness fixes for the SMT backend

### DIFF
--- a/src/smt/preprocess.ml
+++ b/src/smt/preprocess.ml
@@ -166,6 +166,7 @@ let make_salt_explicit sq =
       [<<a_1, ..., a_n>> = <<b_1, ..., b_n>>]  -->
       [a_1 = b_1 /\ ... /\ a_n = b_n]
       ```
+    - Rewrite `[x = TRUE]  -->  [x <=> TRUE]`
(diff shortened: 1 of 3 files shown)


src/smt/preprocess.ml
```diff
@@ -187,10 +186,6 @@ let prepreproc sq =
           let hs,_ = List.split res in
           Smt.record_signatures := hs :: !Smt.record_signatures;
           Record (List.map (fun (s, e) -> (s, self#expr scx e)) res) @@ e
-      | Apply ({core = Internal B.Eq}, [x ; {core = Internal (B.TRUE | B.FALSE)} as y])
```
muenchnerkindl, line 190: I believe you disabled this rewrite because `=' and `<=>' have different strength and so there could be an unsoundness if the rewrite was applied to formulas with unsuitable polarity? Would you mind leaving a comment in the code just as you explained the removal of the rewrite about CHOOSE?

damiendoligez (approved):
LGTM

ahelwer:
@damiendoligez since this is causing a failing test can it be reverted so that the main branch has all tests pass in the CI? It can be merged again when it doesn't cause a test to fail.

kape1395:
It would be better to merge #98 instead of reverting this PR.

PR #57, opened 2022-01-10
_API: mainly related to expressions

johnyf:
- refactoring of internal interfaces, mainly related to expressions
- changes to error messages
- change of directory name to `tla_mode`


src/proof/p_visit.ml
```diff
@@ -168,13 +168,16 @@ class virtual ['hyp] map = object (self : 'self)
...
+            begin
+            let opaque = Opaque v @@ dw in
+            let opaque = self#expr scx opaque in
+            match opaque.core with
+            | Opaque _ -> [Dvar v @@ dw]
+            | _ -> Errors.bug ~at:dw
+                ("unexpected result from `self#expr` \
+                for the identifier `" ^ v ^
+                "` in a `BY DEF`")
+            end
```
damiendoligez, line 180: Could you explain this change? I see that the old code was moved to method `usable` in `p_anon.ml` but how is it called and why do we need this new code in addition?

damiendoligez (commented):
Looks good except for one part that I don't understand.

src/proof/p_visit.ml
damiendoligez: Thanks for the explanation.

damiendoligez (approved):
Looks good now that I understand this opaque business.

PR #58, opened 2022-01-10
REF: add functions for creating syntax-tree nodes, revise internal interfaces

johnyf:
- refactoring that revises internal interfaces related to the syntax tree
- adds new functions for creating syntax-tree nodes


damiendoligez (approved):
LGTM. There are nice reductions of duplicated code here.

PR #61, opened 2022-01-10
formatting of modules related to indexing and subexpression references

johnyf:
Formatting of modules that convert operator references, and expand subexpression references.


damiendoligez:
This PR is only about reformatting some of the source code. I don't think it's worth the risk of introducing merge conflicts with other PRs. Closing without merging. Raise your hand if you disagree.

kape1395:
After the majority of existing branches are merged, it would be nice to reformat all the code using `dune fmt`.

PR #62, opened 2022-01-12
Examples involving use of WF.

kape1395:
If you consider these examples useful, I could add more comments to them.

### DIFF
--- a/examples-draft/SimpleExampleWF_Flag.tla
+++ b/examples-draft/SimpleExampleWF_Flag.tla
@@ -1,4 +1,22 @@
 -------------------- MODULE SimpleExampleWF_Flag --------------------
+(*
+This is a trivial example, showing how to use Weak Fairness in
+proofs. Lets assume we have a flag x, that is off, and later it
+becomes on (e.g. some decision is made). We want to prove, that
+this is eventually done. The general scheme here is to show that:
+
+  - always either we have <>x or <><<Next>>_x;
+  - moreover if we execute <<Next>>_x, x becomes on.
+  - From this it follows, that <>x is satisfied.
+
+To prove <>x \/ <><<Next>>_x we have to consider the fairness
+condition. More details on this are provided in the proof.
+This proof is simple in this case because x \in BOOLEAN, thus
+either the target is reached, or the Next action is enabled.
+
+The module SimpleExampleWF_Max provides an example, where an
+induction is required to show P \/ ENABLED <<A>>_vars.
+*)
 EXTENDS TLAPS
 VARIABLE x
 Init == x = FALSE
@@ -16,9 +34,25 @@ LEMMA SpecTypeOK == Spec => []TypeOK
 LeadsToX == <>x
 THEOREM SpecLeadsToX == Spec => LeadsToX
   <1>1. Spec => [](<>x \/ <><<Next>>_x)
+        (*
+        Since x \in BOOLEAN, we either have x, or <<Next>>_x is
+        enabled because it is precondition is ~x. That simplifies
+        this example a lot. The example with the Max function has
+        a non-boolean variable, thus the proof becomes more involved.
+        *)
         <2>1. x \/ ENABLED <<Next>>_x BY ExpandENABLED DEF Next
-        <2>2. WF_x(Next) => (<>x \/ <><<Next>>_x) BY <2>1, PTL
-        <2>3. QED BY <2>1, <2>2, PTL DEF Spec, Live
+        (*
+        The next statement follows trivially, because
+        WF_x(Next) = []( ([] ENABLED <<Next>>_x) => <><<Next>>_x)
+                   = [](~([] ENABLED <<Next>>_x) \/ <><<Next>>_x) 
+                   = []( (<>~ENABLED <<Next>>_x) \/ <><<Next>>_x)
+                   = []( (<>x                  ) \/ <><<Next>>_x)
+        *)
+        <2>2. WF_x(Next) => [](<>x \/ <><<Next>>_x) BY <2>1, PTL
+        (*
+        The Spec implies WF_x(Next), thus the goal is proven. 
+        *)
+        <2>3. QED BY <2>2 DEF Spec, Live
   <1>2. ASSUME TypeOK, ~x, <<Next>>_x PROVE x' BY <1>2 DEF Next
   <1>3. QED BY <1>1, <1>2, SpecTypeOK, PTL DEF LeadsToX
 
@@ -33,5 +67,5 @@ THEOREM SpecLeadsToStableX == Spec => LeadsToStableX
   
 =============================================================================
 \* Modification History
-\* Last modified Wed Jan 12 16:11:04 EET 2022 by karolis
+\* Last modified Thu Jan 13 00:16:19 EET 2022 by karolis
 \* Created Sat Jan 08 20:57:08 EET 2022 by karolis
--- a/examples-draft/SimpleExampleWF_Flag.tla
+++ b/examples-draft/SimpleExampleWF_Flag.tla
@@ -60,12 +60,11 @@ THEOREM SpecLeadsToX == Spec => LeadsToX
 LeadsToStableX == <>[]x
 THEOREM SpecLeadsToStableX == Spec => LeadsToStableX
   <1>1. Spec => [](x => []x)
-        <2>1. Init => (x => []x) BY DEF Init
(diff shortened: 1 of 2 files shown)


examples-draft/SimpleExampleWF_Max.tla
```diff
@@ -0,0 +1,213 @@
...
+        BY DEF m!Init, m!TypeOK
+  <1>2. m!TypeOK /\ [m!Next]_mX => m!TypeOK'
+        BY m!Assms DEF m!TypeOK, m!Next
+  <1>q. QED BY <1>1, <1>2, PTL DEF m!Spec
+
+(*
+Helper invariant: mIsMax is stable.
+*)
+LEMMA mSpecStableMax == m!Spec => [](mIsMax => []mIsMax)
+  <1>1. m!Init /\ mIsMax => mIsMax
```
muenchnerkindl, line 72: This step is actually unnecessary.

examples-draft/SimpleExampleWF_Max.tla
```diff
@@ -0,0 +1,213 @@
...
+
+Max(a, b) ==
+  IF a > b THEN a ELSE b 
+
+MaxProps == \A a, b \in Nat:
+              /\ Max(a, b) >= a
+              /\ Max(a, b) >= b
+              /\ Max(a, b) \in Nat
+              /\ Max(a, b) \in {a, b}
+THEOREM MaxIsGEQ == MaxProps
```
muenchnerkindl, line 28: Wouldn't it be easier to get rid of one of the names MaxProps or MaxIsGEQ and just state the theorem?

examples-draft/SimpleExampleWF_Max.tla
```diff
@@ -0,0 +1,213 @@
...
+              BY <2>1, <2>2, m!Assms
+              DEF mIsMax, Max, m!Next, mXInv
+        <2>3. CASE mX # m!Null /\ mIsMax
+              BY <2>1, <2>3, m!Assms
+              DEF mIsMax, Max, m!Next, mXInv
+        <2>4. QED BY <2>1, <2>2, <2>3, <1>0, <1>x, m!Assms DEF mXInv
+  <1>q. QED BY <1>1, <1>2, PTL DEF m!Spec
+
+(*
+Prove the that MaxAsFSM produces the same results as Max eventually.
```
muenchnerkindl, line 108: "the that"

examples-draft/SimpleExampleWF_Max.tla
```diff
@@ -0,0 +1,213 @@
...
+CONSTANTS a, b
+ASSUME Assms == a \in Nat /\ b \in Nat
+VARIABLES x
+Null == CHOOSE n : n \notin Nat
+Init == x = Null
+Next ==
+ /\ x = Null
+ /\ \/ a >  b /\ x' = a
+    \/ a =< b /\ x' = b
+Live == WF_x(Next)
```
muenchnerkindl, line 44: A suggestion: instead of baking the proofs of the ENABLED condition into the proofs below, why not prove

LEMMA EnabledNext == x = Null => ENABLED <<Next>>_x

(you may also prove the equivalence, but this is the implication that you need).
In order to do that, show that

Next => << Next >>_x

then use ENABLEDrules to reduce the proof of the assertion to that of

x = Null => ENABLED Next

which (I believe) should be proved by ENABLEDrewrite.

examples-draft/SimpleExampleWF_Max.tla
```diff
@@ -0,0 +1,213 @@
...
+                      = [](~([] ENABLED <<m!Next>>_x) \/ <><<m!Next>>_x) 
+                      = []( (<>~ENABLED <<m!Next>>_x) \/ <><<m!Next>>_x)
+                      = []( (<>mIsMax               ) \/ <><<m!Next>>_x)
+
+        From this and the fact that Spec implies WF, we conclude <1>2.
+        *)
+        <2>1. m!Spec => []P
+              <3>1. m!Init => P
+                    <4>1. SUFFICES ASSUME m!Init PROVE P OBVIOUS
+                    <4>2. SUFFICES ASSUME TRUE PROVE ENABLED <<m!Next>>_mX BY DEF P
```
muenchnerkindl, line 139: These two steps could obviously be combined into

<4>1. SUFFICES ASSUME m!Init PROVE ENABLED <<m!Next>>_mX  BY DEF P

examples-draft/SimpleExampleWF_Flag.tla
```diff
@@ -0,0 +1,71 @@
...
+        *)
+        <2>3. QED BY <2>2 DEF Spec, Live
+  <1>2. ASSUME TypeOK, ~x, <<Next>>_x PROVE x' BY <1>2 DEF Next
+  <1>3. QED BY <1>1, <1>2, SpecTypeOK, PTL DEF LeadsToX
+
+
+LeadsToStableX == <>[]x
+THEOREM SpecLeadsToStableX == Spec => LeadsToStableX
+  <1>1. Spec => [](x => []x)
+        <2>1. Init => (x => []x) BY DEF Init
```
muenchnerkindl, line 63: This step appears to be unnecessary.

muenchnerkindl (commented):
Hello, thanks a lot for your examples. I left a few comments in the modules about how I believe they could be simplified a little.
Best regards, Stephan

kape1395 (commented):
Thanks for the comments!
Tried to consider them and update the spec accordingly, but I got few question (in comments).

examples-draft/SimpleExampleWF_Max.tla
kape1395: I'm sorry for sending the above comment to early. Applying `ExpandENABLED` and adding `mX = mX'` made the step passing.

Nevertheless it would be really useful if you could point to the relevant parts in the source code to understand ENABLEDrewrites and ENABLEDrules and related directives.

examples-draft/SimpleExampleWF_Max.tla
```diff
@@ -0,0 +1,197 @@
...
+VARIABLE mX
+ASSUME mAssms == mA \in Nat /\ mB \in Nat
+m == INSTANCE MaxAsFSM WITH
+  a <- mA,
+  b <- mB,
+  x <- mX
+
+mIsMax == mX = Max(mA, mB)
+
+LEMMA mEnabledNext == mX = m!Null => ENABLED <<m!Next>>_mX
```
kape1395, line 50: @muenchnerkindl 
Was playing with this proof to understand why it works. Found a case making the TLAPS to hang:
```
LEMMA mEnabledNext == mX = m!Null => ENABLED <<m!Next>>_mX
  (* First we show that stuttering steps are not allowed by the action. *)
  <1>1. m!Next => <<m!Next>>_mX
        BY m!Assms DEF m!Next, m!Null
  (* This allows us to drop ... /\ mX # mX' in the goal. *)
  <1>2. SUFFICES ASSUME mX = m!Null PROVE ENABLED (m!Next /\ mX' = mA)
        BY <1>1, m!Assms, ExpandENABLED DEF m!Next, m!Null
  (* Finally by replacing all the equations involving primed variables
     with TRUE (by ENABLEDrewrites) we prove our goal. *)
  <1>4. QED BY <1>2, ENABLEDrewrites, m!Assms DEF m!Next
```
Notice the `/\ mX' = mA` in <1>2. The expression itself is not meaningful in this context, but it breaks the prover somehow.
The TLAPS doesn't crash, but hangs (was waiting for several minutes) and consumes CPU with no visible progress.

muenchnerkindl (commented):
Thank you for your feedback. Maybe I didn't make it clear enough that my comments were just suggestions, feel free to adopt what you feel relevant and ignore the rest.

I believe that @johnyf has some more feedback so I'll wait for his comments before merging your PR.

Regards,
Stephan

johnyf (changes requested):
Thank you for the examples. Please find comments about the code above. The branch `updated_enabled_cdot` is in the process of being rebased. When the changes of the branch `updated_enabled_cdot` have been merged into branch `main`, this pull request can be rebased on branch `main`, before merging. In that rebase, the commits can be squashed into one commit that adds the new files.

damiendoligez:
@kape1395 @muenchnerkindl
Do you think we can merge this PR as is, or do you want to work on it some more?

kape1395:
I'm unsure. Maybe there are enough other examples on WF, and this could be dropped.

PR #68, opened 2022-06-20
CI: show logs for each test file

johnyf:
- BUG: correct quoting of environment-variable values in the script `setup_shell_env.py`, to avoid errors in CI workflow of `release.yml`.

- run workflow of file `.github/workflows/release.yml` whenever the file changes, and also twice a year.

- add underscores and "UTC" to tag name created by GitHub Actions when running `release.yml`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,13 @@ jobs:
                 TEST_RESULT=$?
                 cat _build/default/test/tests.log
                 exit $TEST_RESULT
+            - name: show logs for each test file
+              if: failure()
+              run: |
+                find test -type f \
+                    -name '*.err' -o \
+                    -name '*.out' \
+                    -exec cat '{}' \;
             - name: Clone tlaplus/examples
               uses: actions/checkout@v4
               with:


ahelwer:
@johnyf could you rebase these on the current CI files to see whether these changes should be merged?

ahelwer:
1. Was the original fix for quoting environment variables no longer necessary?
2. I like the practice of dumping test output files to console so we get the output in the CI!
3. I am not sure about the updated release practice; may have to defer to someone else in their thoughts here.

ahelwer:
In #92 there was discussion of moving to a rolling release model (which I have really been dragging my feet on getting around to doing). So I think people want to keep the current release model then transition to that one, instead of releasing every six months + whenever release.yml is modified. However, I like all the other changes!

ahelwer:
Also @damiendoligez as FYI, I believe this is the issue you mentioned in the last community meeting

ahelwer:
@johnyf we implemented rolling releases in #182 so there is no need for the changes to the release CI file. However, I think dumping the test output files is still a good idea; if you'd like to remove the changes to release.yml I will merge this.

Issue #80, opened 2023-05-26
Use Dune for build?


kape1395:
Could a PR be accepted if I manage to add Dune based build to the repository?
I hope that would make the development more accessible; also, it seems the VSCode OCaml extension depends on it.
Maybe this repo avoids using this build system intentionally?

PR #81, opened 2023-05-26
Basic support for build with dune.

kape1395:
Closes https://github.com/tlaplus/tlapm/issues/80

This build is useful at least to have the `vscode` IDE working for the `tlapm` codebase.

Additionally, for some reason, the dune build was failing on `src/backend/fpfile.ml` while `make` build passed. Looking at the code, it seems to me like the mistake actually exists. The fixed version compiles with the `make` build as well. Would be glad if someone would take a look at this and explain, why the original build was not failing.


damiendoligez (approved):
LGTM.

Issue #82, opened 2023-05-28
Simplify the build process.


kape1395:
I would like to reorganize the build/release process for TLAPM. It was always very complicated for me to understand how to build/test some changes or debug system errors in TLAPM. I plan to work on the VSCode Tlapm support, and some changes to the Tlaps might be needed (moving some functions to Tlapm, which are now implemented in the toolbox directly). Therefore I try to make the development of Tlapm a bit more comfortable.

From my limited knowledge, the dune build system is widely used for OCaml projects, so maybe we should try it out. I have added basic support for it in https://github.com/tlaplus/tlapm/pull/81, but it only covers what is needed to work with the code base in VSCode. I would like to reorganize the build further to be able to run tests at least. The following are the main concerns for me so far:

  - If I understood correctly, the current tests are tied to the release process. One has to make a release, then install it, and the run the test case with it. In typical scenarios, one runs tests before making a release. Also, building/installing a release make the development cycle slower. I would like to set up the dependencies (Isabelle, Zenon, ...) during the build process, so they would be available for tests.
  - The current release builds the TLA+ Isabelle heap on installation. Could the TLAPS be shipped with the heap already pre-built?
    Locally I updated the build scripts to download the Isabelle during the build process (instead of the release script) and build the TLA+ Theory heap. I put the heap into the Isabelle/heaps/*/TLA+ so the installation does not depend on the user's environment, so that's doable. The same Isabelle could go as part of the release.
  - The current `./configure` script doesn't check a lot. Maybe we can use a dune-based build instead of autoconf/automake/make? That would make the build/test/release process similar to other OCaml projects.
  - The current release file is an executable that extracts the attached files and builds the needed Isabelle theories. If we make the Isabelle shipped with the TLA+ heap pre-built, it would be enough to have a tar.gz as a release, which would be enough to extract. I think it would be easier to use (it's always scary to run downloaded scripts/installers). Such change would also make it easier in the future to provide Debian packages (e.g., via PPA) to make installation more convenient.
  - There is a patch now applied when building LS4. Having a fork of LS4 in GitHub with a branch containing that change would be easier. We could then use it as a git sub-module or clone it during the build process. Again, that would make the LS4 available for the tests without building and installing a release.

If you have no big objections to this direction, I could make a PR.

PR #83, opened 2023-05-31
Dune based build

kape1395:
Closes https://github.com/tlaplus/tlapm/issues/82

This PR extends the dune-based build started at https://github.com/tlaplus/tlapm/pull/81.
This version supports running the tests and creating a release.

The changes are non-invasive, so the previous build/test process remains working.


damiendoligez (commented):
Looks good except for the change to `ocaml_versions.py`.

I confirm that the change to `fpfile.ml` fixes a bug.

kape1395:
The PR is ready to be reviewed. In general, it adds:

  - Make the build process familiar to other projects (`make ; make test; make install; make release`).
  - The distribution file is just an archive that is enough to extract to use (no installation).
  - The test cases can be run without making a release and installing it. Use `make -C test -f Makefile.dune fast` or similar. The shell autocompletion to select a more detailed test case works this way.
  - Tlaps, if built with dune, uses dune sites to locate the backend provers and stdlib tla files. This makes taps easier to use, as no PATH updates are needed. As a side-effect, other projects can supply their files to these locations (sites); thus, a plugin system is in place to use, e.g., to add additional tla libraries.

There are some problems still, but I hope those are not blockers:
  - I found no simple way to run unit (kaputt-based, provided in the mlt files) tests without breaking the existing build. There are a few of them; maybe I could rewrite them using `ppx_inline_test` or similar?
  - The current dune install step breaks the symlinks and exec permissions. As a workaround, I fix them afterward with a post-install script. I will try to make a fix/PR to the dune project, but its acceptance will not be fast, I guess. Then we could remove that workaround with the post-install script.
  - There are now two parallel build systems. Would you agree to remove the old one and keep the dune based only?
  - I cannot test the scripts on operating systems other than Linux. I would like to have help or pointers with this.

deps/isabelle/Makefile
```diff
@@ -0,0 +1,55 @@
+ISABELLE_VSN:=Isabelle2011-1
+ISABELLE_TGZ:=$(ISABELLE_VSN)_bundle_x86-linux.tar.gz
```
damiendoligez, line 2: This fails in MacOS, and most probably on Windows.

In MacOS, you have to download a `.dmg` file, which is not a `.tar.gz`, and use a different set of commands to extract files from it. IOW, the recipe for `$(ISABELLE_DIR): $(ISABELLE_TGZ)` will have to be completely different. Does `dune` give you access to the OS name, or will you need to make this a `Makefile.in` and tweak `configure`?

deps/z3/Makefile
```diff
@@ -0,0 +1,20 @@
+Z3_VERS=4.8.9
+Z3_URL_PREFIX=https://github.com/Z3Prover/z3/releases/download/z3-$(Z3_VERS)
+Z3_BASE_NAME=z3-${Z3_VERS}-x64-ubuntu-16.04
```
damiendoligez, line 3: Like Isabelle, this is not going to work on MacOS and Windows.

src/dune_site_mock/setup_paths.ml
```diff
@@ -0,0 +1,9 @@
+(*
+This file is used as a mock for the Setup_paths module generated by dune-site during the dune-based build.
+In the make-based setup we will retain the approach to find the backend proovers via the PATH environment variable.
+Hence this file inicates (by returning empty list) the absense of the dune-site-based paths, thus a fallback to PATH.
```
damiendoligez, line 4:
```suggestion
In the make-based setup we will retain the approach of finding the backend provers via the PATH environment variable.
Hence this file indicates (by returning empty list) the absense of the dune-site-based paths, thus a fallback to PATH.
```

src/params.ml
```diff
@@ -68,12 +68,26 @@ type executable =
...
+    let auxdir = Filename.concat library_path "bin" in
+    let extrapath = sprintf ":%s:%s" mydir auxdir in
+    let path = Sys.getenv "PATH" in
+      sprintf "%s%s" path extrapath
+  | backends_site ->
+    let site_bin bs = Filename.concat bs "bin" in
+    let site_isa bs = Filename.concat (Filename.concat bs "Isabelle") "bin" in
+    let site_paths bs = [site_bin bs; site_isa bs] in
+    let path_elems = List.concat (List.map site_paths backends_site) in
+      sprintf "%s:%s" (String.concat ":" path_elems) (Sys.getenv "PATH")
```
damiendoligez, line 87: This is a pretty big change: you're putting the extra path elements at the beginning of PATH rather than the end.

src/params.ml
```diff
@@ -239,7 +253,10 @@ let set_smt_logic logic = smt_logic := logic
 
 let max_threads = ref nprocs
 
-let rev_search_path = ref [library_path]
+(* The actual list if paths at which the library TLA files are searched. *)
```
damiendoligez, line 256:
```suggestion
(* The actual list of paths at which the library TLA files are searched. *)
```

tlapm.opam
```diff
@@ -0,0 +1,50 @@
...
+]
+authors: [
+  "Kaustuv Chaudhuri"
+  "Denis Cousineau"
+  "Damien Doligez"
+  "Leslie Lamport"
+  "Tomer Libal"
+  "Stephan Merz"
+  "Jean-Baptiste Tristan"
+  "Hernan Vanzetto"
```
damiendoligez, line 23: again, needs updating

zenon/regression/examples/data/TLAPS.tla
```diff
@@ -1 +1 @@
-../TLAPS.tla
\ No newline at end of file
+../../../../library/TLAPS.tla
```
damiendoligez, line 1: I think this symlink should be removed rather than updated.

src/dune_site_mock/setup_paths.ml
kape1395: The entire file is now removed. It was needed only to support two build systems in parallel.

zenon/regression/examples/data/TLAPS.tla
kape1395: Removed it.

test/TOOLS/do_junit_tests
```diff
@@ -3,7 +3,20 @@
 
 BASEDIR="`pwd`"
 TOOLDIR="$BASEDIR/TOOLS"
-export TLAPM="pkill -f cvc3 | $BASEDIR/../tlapm --stretch 10 --threads 4 -I $BASEDIR/../library"
+
+if [ -z "$USE_TLAPM" ]; then
+  TLAPM_BIN=$BASEDIR/../tlapm
+else
+  TLAPM_BIN="`realpath $USE_TLAPM`"
```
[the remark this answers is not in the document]
kape1395: I have updated the scripts to avoid using `realpath`.

deps/z3/Makefile
kape1395: I have updated the makefiles (for Isabelle and other backends). All work on MacOS and Cygwin except Isabelle:
  - On MacOS, I'm getting `poly: Bad CPU type in executable`, I guess the 32 bit support is missing. At least it was needed on Linux for Isabelle to work.
  - On Cygwin, the following DLLs/Versions are missing: `cyggcc_s-1.dll`, `cyggmp-3.dll`. The latter was added in the old release script, but the DLL is not committed to the repo, thus I guess it should fail as well.

deps/isabelle/Makefile
kape1395: I have updated the makefile to consider different OS types. Isabelle still doesn't work.

  - On MacOS, I'm getting `poly: Bad CPU type in executable`, I guess the 32 bit support is missing. At least it was needed on Linux for Isabelle to work.
  - On Cygwin, the following DLLs/Versions are missing: `cyggcc_s-1.dll`, `cyggmp-3.dll` and I can't find them for installation in the Cygwin package repository. The old release script was adding `cyggmp-3.dll` to the Isabelle, but the DLL is not committed to the repo; thus, I guess it should fail as well.

tlapm.opam
kape1395: This file is generated from `dune-project`; it will be enough to update these lists in a single place.

src/params.ml
kape1395: My motivation for this change is to have the backends defined more predictably. E.g., I have an existing installation of TLAPM, and its backends are on the path. It is easy to accidentally forget to set the path to the new location and invoke a new TLAPM with old backends.

Behavior may be more predictable if TLAPM always used the backends shipped with it. We could introduce options to override specific backends, e.g., `--with-z3=/home/.../bin/z3` and similar? Of course, an alternative to use environment variables could also be supported, e.g., `export TLAPM_Z3=/home/.../bin/z3`.

Some setups probably exploit the current implementation to look for backends in the PATH variable to override the backends. If the above proposal doesn't look good, I can put these paths at the end. This will also require updating the tests to avoid accidental overrides.

kape1395:
I have resolved `poly: Bad CPU type in executable` for MacOS and several other incompatibilities in makefiles.

kape1395:
@damiendoligez, could you rerun the pipeline? I hope I have resolved the problems with MacOS.

src/params.ml
damiendoligez: Coming back to this, I now think you're right: we should make it hard to accidentally override the provided backends. So let's keep this change and we'll add an override mechanism later if needed.

damiendoligez:
@kape1395 I don't have any button on my GitHub interface to trigger the CI checks.

damiendoligez (approved):
I've reviewed all the files that changed since my last review and everything looks good.

kape1395:
> I've reviewed all the files that changed since my last review and everything looks good.
> 

The checks are now executed. Maybe I have mistakenly assumed it is manual.

kape1395:
@johnyf, @damiendoligez, maybe this PR can be merged already?

kape1395:
@ahelwer, this is the PR I have updated the CI scripts.
There is room for improvement, especially when a release policy is decided.

ahelwer:
@kape1395 okay cool, is there any obstacle to merging the PR? As you said it would be nice to get it in so we don't have PRs piling up on top of PRs. Don't have to issue a release yet.

ahelwer:
Hurrah it's merged! @kape1395 do you think this is a good time to add a CI step to test the newly-built TLAPM against the tlaplus/examples repo, like we do for the tools here? https://github.com/tlaplus/tlaplus/blob/dd9f38f26c184014086c37bff8a056b4d51f4464/.github/workflows/pr.yml#L54-L83

kape1395:
@ahelwer, that's probably a good time from the build-script changes perspective, but now #55 was also merged to the main branch and introduced a bug. One of the tests fails; it sounds related to CHOOSE handling or theorem expansion. I hope to look at it deeper next weekend.

kape1395:
@ahelwer, I have updated the build/install instructions with this PR. Probably that can be useful for you: https://github.com/tlaplus/tlapm/blob/main/INSTALL.md

ahelwer:
@kape1395 okay cool! Once the new bug is fixed do you think it will be worth cutting a pre-release so we can start consuming it in the tlaplus/examples CI to test that the nondeterministic install bug was fixed?

kape1395:
This question is not to me probably. What do you think about using the latest main all the time? This way some kind of continuous integration would be implemented.

lemmy:
We used to have that and I'm very much in favor or adding it again. :-)

ahelwer:
I'd say that depends how much effort is necessary to build TLAPM from a checkout of main. The tlaplus tools are pushed to a nightly build server every time a commit hits main, which is nice, and we pull that down in the tlaplus/examples CI. Does this repo do anything similar?

kape1395:
What OS are you using locally? That could be some incompatibility in the CLI tools.
That line is generated in `deps/isabelle/Makefile` by the line
```
echo "all:\n\t chmod +x \$$(FILES)" >> $@
```

I guess in your OS, `echo` requires some flags (e.g., `-e`) to process escaped symbols properly.

Can you try changing that line to 
```
echo -e "all:\n\t chmod +x \$$(FILES)" >> $@
```

kape1395:
Nah, it fails on my machine with -e added. Maybe there is a mismatch with the make versions? In my case that's
```
$ make --version
GNU Make 4.3
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

ahelwer:
Hi @kape1395 I can confirm adding the `-e` flag to `echo` fixed the issue for me. Now I just have some tests failing:
```
cd test && env \
        USE_TLAPM=../_build/tlaps-release-dir/bin/tlapm \
        USE_LIB=../_build/tlaps-release-dir/lib/tlapm/stdlib \
        ./TOOLS/do_tests fast/basic
  fast/basic/cvc3_false_test.tla FAILED
  fast/basic/isa_false_test.tla FAILED
  fast/basic/isa_true_test.tla FAILED
  fast/basic/nonexistent_test.tla FAILED
  fast/basic/smt_arith_test.tla FAILED
  fast/basic/smt_false_test.tla FAILED
  fast/basic/tab_character_test.tla FAILED
  fast/basic/true_test.tla FAILED
  fast/basic/z3_arith_test.tla FAILED
  fast/basic/z3_false_test.tla FAILED
  fast/basic/zen_false_test.tla FAILED
  fast/basic/zen_true_test.tla FAILED

total:   12
success: 0
failure: 12

**** WARNING: SOME TESTS FAILED ****

make: *** [Makefile:51: release] Error 3
```
My version of make is 4.4.1.

kape1395:
You can check the error in files e.g.
```
./_build/default/test/fast/basic/cvc3_false_test.tla.err
./_build/default/test/fast/basic/cvc3_false_test.tla.out
```
and correspondingly for other test case files.

Adding `-e` is probably not enough, as that makes the script non-portable. For example, it fails on my machine. I will experiment with that.

By the way, why do you need the bin installer? There is no installer anymore.
After running
```
make build install
```
you can run tlapm e.g. as `opam exec -- tlapm`.
Or that's not enough to run the tests?
In my case, the tlapm exec is installed then as `~/.opam/5.1.0/lib/tlapm`, but I think using `opam exec` is probably more convenient.

PR #99, opened 2023-11-07
fix install location for ptl_to_trp

damiendoligez:
`ptl_to_trp` is installed in `.../bin` but `tlapm` expects to find it in `.../lib/tlapm/backends/bin`.


kape1395 (approved):
That looks right to me.

The previous version was working for me probably because I was running `make install` before running the tests. This way, `ptl_to_trp` appeared in the `$PATH`, and the current tlapm implementation uses $PATH as a fallback.

Issue #133, opened 2024-05-29
ISA: Example in `isabelle/examples/AtomicBakeryG.thy` fails.


kape1395:
This example fails after upgrading to Isabelle2024.
It might be because of the Nat/Int rewrite, but I'm unsure.

kape1395:
The last commit this example was working is eed59fc6db0f99b14a1860b9ddf2580eaebf3415,
It passes there with `Isabelle2024`,  and a minor fix in the ROOT file
```
$ git diff
diff --git a/isabelle/examples/ROOT b/isabelle/examples/ROOT
index cbcd3fa..8033e83 100644
--- a/isabelle/examples/ROOT
+++ b/isabelle/examples/ROOT
@@ -6,7 +6,7 @@ Examples of using Isabelle/TLA+
 *)
 
 (* build the session (without a heap) using "isabelle build -D ." *)
-chapter "TLA+ Examples"
+chapter "TLA+Examples"
 session "TLA-Examples" = "TLA+" +
   options [document=pdf, document_output = "output", document_variants="document:outline=/proof"]
```
Tried to build it as follows in the `isabelle/examples` folder
```
~/Programs/Isabelle2024/bin/isabelle build -d ../ -D . -f -c
```

The next commit touching `/isabelle/` is d44f1a1c64844a5bce80af73b1717a85c1593bdd; the example fails here.

@muenchnerkindl, maybe that's useful for you.

kape1395:
Resolved in #109 and #134 .

PR #134, opened 2024-05-31
Rewrite Nat to {x ∈ Int : 0 ≤ x} instead of n ∈ Nat to (n ∈ Int ∧ 0 ≤ n).

kape1395:
This PR tries to address https://github.com/tlaplus/tlapm/issues/133.

The idea is to rewrite `Nat` as a value instead of its use in expressions like `n ∈ Nat`. So expressions like `[P -> Nat]` are also rewritten. E.g. the following lemma was failing before this change (while the same with `a` instead of `Nat` was passing):
```
lemma
  fixes p :: "c" and f and i
  assumes "f ∈ [p → Nat]" and "i ∈ p"
  shows "f[i] ∈ Nat"
  using assms by auto
```

All the tests are still passing (tests, community modules, examples repo).
The `isabelle/examples/AtomicBakeryG.thy` example still fails, but now the first error is at line `535` instead of `325`.

### DIFF
--- a/isabelle/ROOT
+++ b/isabelle/ROOT
@@ -10,7 +10,7 @@ session "TLA+" = "Pure" +
 
   document_files
     "root.tex"
-(*
+
 session "TLA+Tests" in tests = "TLA+" +
   options [document=false]
   theories
@@ -21,8 +21,7 @@ session "TLA+Examples" in examples = "TLA+" +
 
   theories
     Allocator
-    (* AtomicBakeryG  TODO: Restore it here, when examples are fixed. *)
+    AtomicBakeryG
 
   document_files
     "root.tex"
-*)
--- a/isabelle/ROOT
+++ b/isabelle/ROOT
@@ -10,7 +10,7 @@ session "TLA+" = "Pure" +
 
   document_files
     "root.tex"
-
+(*
 session "TLA+Tests" in tests = "TLA+" +
   options [document=false]
   theories
@@ -25,3 +25,4 @@ session "TLA+Examples" in examples = "TLA+" +
 
   document_files
     "root.tex"
+*)
--- a/isabelle/examples/Allocator.thy
+++ b/isabelle/examples/Allocator.thy
@@ -8,7 +8,7 @@
 section \<open>A Simple Resource Allocator\<close>
 
 theory Allocator
-imports "../Constant"
+imports Constant
 begin
 
 text \<open>
--- a/isabelle/examples/Allocator.thy
+++ b/isabelle/examples/Allocator.thy
@@ -2,13 +2,13 @@
     Author:     Stephan Merz, Inria Nancy
     Copyright (C) 2009-2022  INRIA and Microsoft Corporation
     License:    BSD
-    Version:    Isabelle2021-1
+    Version:    Isabelle2024
 *)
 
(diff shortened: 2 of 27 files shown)


muenchnerkindl:
@kape1395 I went over your changes and made some minor changes. For example, I commented out lemma ```bspec' ''' in SetTheory because it didn't seem to make any difference anymore. Also, I fixed example ```AtomicBakeryG''', which pointed to some missing rewrite rules for symmetric cases etc. Not terribly proud of it (it is slow, and it was never pretty), but at least it works again. At some steps, I had to add ```bspec''' as a safe elimination rule (```elim!''') to make the proof work, which seems to indicate that some more optimization is needed in this respect.

isabelle/ROOT
```diff
@@ -10,7 +10,7 @@ session "TLA+" = "Pure" +
 
   document_files
     "root.tex"
-
+(*
```
kape1395, line 13: @muenchnerkindl, can I uncomment this?
I usually open the theories as `isabelle jedit -D isabelle/ isabelle/Constant.thy`, then all the modules see each other. I use these additional sessions in the ROOT file to build the test/examples theories during the tests.

isabelle/examples/Allocator.thy
```diff
@@ -2,13 +2,13 @@
...
     License:    BSD
-    Version:    Isabelle2021-1
+    Version:    Isabelle2024
 *)
 
 section \<open>A Simple Resource Allocator\<close>
 
 theory Allocator
-imports Constant
+imports "../Constant"
```
kape1395, line 11: @muenchnerkindl, if the theories are referenced from the same ROOT file, then the other theories are imported without the relative path, just `imports Constant`. Can I revert this change as well?

isabelle/ROOT
muenchnerkindl: Sure, please go ahead. Made the change when I built the heap and forgot to revert it.

isabelle/examples/Allocator.thy
muenchnerkindl: Of course. Made the change in interactive mode when I modified both the theories in the main directory and in the examples directory.

kape1395:
@muenchnerkindl, @damiendoligez, suddenly, I started getting this error on all the Isabelle invocations, even on the branches that were working previously.

https://github.com/polyml/polyml/blob/90c0dbb2514e1498819da4c14bbeff62481e2149/libpolyml/savestate.cpp#L1141

That timestamp check sounds strange to me, but these lines have been here for 15 years already.
Have you seen something similar?

muenchnerkindl:
@kape1395 I have never seen this. There was a message sent to the Isabelle list about two years ago [1], but it was then retracted, saying it had been sent "in error", so not really helpful.

[1] https://mailman46.in.tum.de/pipermail/isabelle-dev/2022-September/017655.html

kape1395:
On the error `The parent for this saved state does not match or has been changed`.
I resolved this by removing `~/.isabelle/` on my machine. Probably that's an interference of interactive sessions with the solver's state.
Maybe I'll try to set `$USER_HOME` to the current directory when invoking the Isabelle backend to avoid such a confusion in the future.

PR #138, opened 2024-06-07
Show ocaml version in https://github.com/tlaplus/tlapm/actions/workflows.

lemmy:
Do not run python to translate an *opaque* array index into a meaningful ocaml version number. Moreover, commit 5e71b24db228d2b7ef4d25e4cbaadb3e80206912 revealed a problem with this; it introduced an out-of-bounds read.

[Build]


kape1395 (approved):
I'm happy to get rid of this mapping.

PR #144, opened 2024-07-14
Emit bytecode in dune build for debugging

ahelwer:
This will generate a `tlapm.bc` OCaml bytecode file in `_build/default/src` which can be loaded by `ocamldebug`. Dune 2.0+ requires this to be explicitly specified ([docs](https://dune.readthedocs.io/en/stable/reference/dune/executable.html)):

>Given an executable stanza with (name \<name\>), Dune will know how to build \<name\>.exe. If requested, it will also know how to build \<name\>.bc and \<name\>.bc.js (Dune 2.0 and up also need specific configuration (see the modes optional field below)).

`ocamldebug` also requires everything to be built with the `-g` flag, which is the case by default with Dune; see output of `dune printenv .`.

Ref https://github.com/tlaplus/tlapm/discussions/143

I also added a debug launch profile for VS Code, and added the [earlybird](https://github.com/hackwaly/ocamlearlybird) DAP server to the `opam-deps-dev` make target.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "TLAPM Debug",
+            "type": "ocaml.earlybird",
+            "request": "launch",
+            "program": "${workspaceFolder}/src/tlapm.bc",
+            "stopOnEntry": true,
+            "env": {
+                "CAML_LD_LIBRARY_PATH": "${workspaceFolder}/_build/default/src:${env:CAML_LD_LIBRARY_PATH}"
+            }
+        }
+    ]
+}
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ opam-deps:
 	opam install ./ --deps-only --yes --working-dir
 
 opam-deps-dev:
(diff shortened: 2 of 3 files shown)


ahelwer:
It should be noted that the generated `tlapm.bc` is very large:

```
>ls -lh
...
-r-xr-xr-x 1 ahelwer ahelwer  22M Jul 14 18:30 tlapm.bc
-r-xr-xr-x 1 ahelwer ahelwer 5.1M Jul 14 18:30 tlapm.exe
...
```
So perhaps these changes must be paired with logic to avoid bundling it into the zip file.

ahelwer:
@kape1395 what do you think of these changes?

kape1395 (approved):
Looks good to me.
As I see, the bc file is not in the release.

ahelwer:
@kape1395 actually I think these changes should be reverted, because after a few weeks playing around with the OCaml debugger I have found it is just too immature to use as a development tool and the utop REPL should be used instead. This seems to be the general idea in the OCaml community as well.

kape1395:
@ahelwer, I would like to keep the changes. Even if they are not that useful today, they don't harm anyone either.

PR #145, opened 2024-07-26
Add license field to opam package definition file

ahelwer:
Without this field, newer versions of `opam` will print out the following warning when running `make opam-deps`:
>[WARNING] Failed checks on tlapm package definition from source at
          git+file:///mnt/data/ahelwer/src/tlaplus/proofs#main:
  warning 68: Missing field 'license'

This tiny change fixes that warning by adding the [SPDX short form](https://spdx.org/licenses/BSD-2-Clause.html) of the repo's license to the `tlapm.opam` file.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -10,6 +10,8 @@
 
 (source (github tlaplus/tlapm))
 
+(license BSD-2-Clause)
+
 (package
  (name tlapm)
  (synopsis "TLA+ Proof Manager")
--- a/tlapm.opam
+++ b/tlapm.opam
@@ -22,6 +22,7 @@ authors: [
   "Jean-Baptiste Tristan"
   "Hernan Vanzetto"
 ]
+license: "BSD-2-Clause"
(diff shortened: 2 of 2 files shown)


kape1395:
This change specified "License: 2-clause BSD", but the README.md mentions

> License: 2-clause BSD, portions under LGPL2.1+LE (see file LICENSE for details)

I checked the https://github.com/tlaplus/tlapm/blob/main/LICENSE, it matches https://opensource.org/license/bsd-2-clause. No exceptions mentioning LGPL are there. Maybe README.md should be updated as well?

@muenchnerkindl, @damiendoligez, maybe you remember why LGPL is mentioned here?

muenchnerkindl:
My guess would be that the `translate' utility (related to the LS4 decision procedure for propositional temporal logic) might originally have been under LGPL, but the README says that the author agreed to let us use it under BSD.
@damiendoligez Do you remember?

kape1395 (approved):

PR #147, opened 2024-08-15
Exposed parse method in TLAPM API

ahelwer:
This adds a simple string -> parse tree method in TLAPM's public API, so it can be hit by non-inline unit tests.

### DIFF
--- a/src/tlapm_lib.ml
+++ b/src/tlapm_lib.ml
@@ -615,7 +615,7 @@ let init () =
        exit 3
 
 (* Access to this function has to be synchronized. *)
-let module_of_string ~(content : string) ~(filename : string) ~loader_paths ~prefer_stdlib : (Module.T.mule, (string option* string)) result =
+let transitive_parse_module_of_string ~(content : string) ~(filename : string) ~loader_paths ~prefer_stdlib : (Module.T.mule, (string option* string)) result =
     let parse_it () =
         Errors.reset ();
         Params.prefer_stdlib := prefer_stdlib;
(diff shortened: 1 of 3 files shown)


kape1395:
I added a similar function in the LSP branch.

https://github.com/kape1395/tlapm/blob/0ac8eac06ccb96a4cb138e622e15fa4aaccf41ac/src/tlapm_lib.ml#L617

Mine does more about related modules. Is is meaningful to keep them both?

ahelwer:
I think so, because it would be nice to just generate a syntax tree without anything else.

kape1395:
@ahelwer, the LSP branch has now been merged.
Can you fix the conflict? What do you think about renaming
  - my function: `module_of_string` -> `elaborate_module_of_string`
  - and your: `parse_module_from_string` -> `parse_module_of_string`?

ahelwer:
How about `transitive_parse_module_of_string` for your method?

kape1395:
I sent you a PR with an attempt to resolve this naming clash.

kape1395:
Why have you reverted `modctx_of_string` to `transitive_parse_module_of_string`. I think the latter is not very precise.

src/tlapm_lib.ml
```diff
@@ -650,4 +650,14 @@ let module_of_string ~(content : string) ~(filename : string) ~loader_paths ~pre
...
          | None, None -> Error (None, Printexc.to_string e))
 
+let module_of_string module_str =
+    let hparse = Tla_parser.P.use M_parser.parse in
+    let (flex, _) = Alexer.lex_string module_str in
+    Tla_parser.P.run hparse ~init:Tla_parser.init ~source:flex
+
 let stdlib_search_paths = Params.stdlib_search_paths
+
+let parse_module_of_string module_str =
```
kape1395, line 660: Is `parse_module_of_string` left here accidentally? It looks the same as `module_of_string`.

kape1395: Cleaned up things after the merge.

kape1395 (approved):

PR #154, opened 2024-09-15
Expose tlapm_lib for other projects.

kape1395:


ahelwer:
@kape1395 Is the idea we would publish TLAPM as a package to OPAM? Do you think it would be a good idea to split off the parser into its own package, since that is probably the only thing people would want to consume?

kape1395:
That would be useful, but that's probably not only the parser.
For example, it was an attempt to use the lib as a TLA+ formatter.

I once tried to make it publishable via OPAM, but the build failed when performed in OPAM while succeeding from the source directly. Something has to be double-checked.

ahelwer:
I think it might not make sense to publish all of TLAPM as an OPAM package, but things like the parser - which could be used in other projects, like a formatter - would make sense. I'll keep the idea of separating the parser out into its own source directory in the back of my mind, I'll be writing a lot of tests for it in the future.

PR #155, opened 2024-09-17
Use cmdliner in translate

glondu:


kape1395 (approved):
I think the change is good. It probably accidentally changed option `-al` to `--al`, but I found no uses for this option in the codebase, so it's probably OK.

PR #159, opened 2024-10-23
Run corpus of TLA+ syntax tests in simple pass/fail manner

ahelwer:
This adds a preexisting set of approximately 200 syntax tests for TLAPM's parser. Future work will build on this to examine the actual parse tree output by TLAPM, but for now we just see whether the parser successfully parses all the things it is expected to. This effort uncovered a fair number of bugs which will be opened as issues in this repository.

Commits were generally made for the purpose of tracking work so I recommend doing a squash-commit if this is merged.

This work was funded by a grant from the TLA+ Foundation.


ahelwer:
@kape1395 this is good to merge. Would appreciate a lookover of my OCaml code, this is my first time writing anything in the language!

PR #177, opened 2024-11-21
Switch tlapm to cmdliner

glondu:
This pull request switches tlapm to cmdliner for command-line parsing. Note that it slightly changes the syntax and reverse dependencies (TLA Toolbox? VSCode?) will have to be fixed.


lemmy:
Unfortunately, there are no resources to change/fix the Toolbox.

ahelwer:
>Note that it slightly changes the syntax and reverse dependencies (TLA Toolbox? VSCode?) will have to be fixed.

Can you expand what is meant by this?

glondu:
> Unfortunately, there are no resources to change/fix the Toolbox.

I will try to fix it myself. What is the proper venue for support? Is there a mailing-list or forum?

ahelwer:
I don't think it is a good idea to break the public API of the command unless there's a very good reason; what do you see as the benefits of this change?

glondu:
> I don't think it is a good idea to break the public API of the command unless there's a very good reason

I generally agree with this statement.

> what do you see as the benefits of this change?

[Cmdliner](https://erratique.ch/software/cmdliner) is a library with a nicer API than Arg that has traction in the OCaml community. It imposes POSIX and GNU conventions on command-line syntaxes and automatically generates manpages.

It turns out that Cmdliner does not (and [won't](https://github.com/dbuenzli/cmdliner/issues/193)) support options with several arguments (as in `--toolbox A B`) because it doesn't fit said conventions. I chose to support `--toolbox A,B` which is Cmdliner's default for this kind of option. Note that `--toolbox=A,B` is also automatically supported. I could have chosen `--toolbox "A B"` as well (but it doesn't solve the compatibility issue).

Note that this pull request doesn't take much advantage Cmdliner's nicer API; in my opinion, more refactoring would be needed. I tried to have the changes localized in a single file for this pull request.

Maybe @muenchnerkindl (the one who asked me to do this) or @damiendoligez have other reasons.

muenchnerkindl:
Before I get too much credit for the idea, the change was suggested by @kape1395 with the objective to ultimately reduce the number of global variables. As @glondu says, the current PR is only a first step towards achieving this goal. Unfortunately, it turned out (much to our surprise) that Cmdliner doesn't support command-line options that take more than one parameter, unless separated by a symbol other than a space.

kape1395:
Maybe for backwards compatibility we can:
  - support `--toolbox A:B` or a variant of that with some delimiters. ":" seems natural to specify a range.
  - If `--toolbox A B` is provided, consider `--toolbox A` as a "named option" and "B" as a separate "positional" argument.
I guess currently, all the positional options are considered files. But if --toolbox only gets one number and there is a positional argument that looks like a number, consider it an end of the range.

That's not very clean from the implementation perspective, but maybe would help to resolve the problem?

kape1395:
The original reason was to reduce the number of global refs. Interpreting the `cmdliner` arguments doesn't sound like hacking it or objecting to the initial goal.

I thought you were in favor of not changing the CLI API:
> I don't think it is a good idea to break the public API of the command unless there's a very good reason; what do you see as the benefits of this change?

ahelwer:
I think breaking CLI API without good reason is bad but greatly simplifying the CLI parsing code (or reducing the number of global refs, although I am not sure what that means) could be a good reason.

I don't think exposing the old CLI API through an extra `--toolbox` flag is a good design; it's easy to update the toolbox to use the new API, the hard part is dealing with people using the old toolbox before those fixes and new version of TLAPM.

damiendoligez:
In https://github.com/glondu/tlapm/pull/1 I propose a small change to patch up the command line before calling Cmdliner.

lemmy:
> I've updated my branch with the tip from Damien which restores the compatibility of two-argument options.

Does this make https://github.com/tlaplus/tlaplus/pull/1085 obsolete?

ahelwer:
Oh wonderful!!! It would be nice to accept & ignore deprecated options if that does not induce any difficulty, although since the next release of TLAPM will probably be at least 1.6.0 (a minor version bump), that does indicate to users that API compatibility breaks are possible so I would accept dropping long-deprecated options.

glondu:
> > I've updated my branch with the tip from Damien which restores the compatibility of two-argument options.
> 
> Does this make [tlaplus/tlaplus#1085](https://github.com/tlaplus/tlaplus/pull/1085) obsolete?

The part about `--toolbox`, yes, but not the part about stopping using deprecated options...

glondu:
> So some old versions of the toolbox still send the deprecated options, which newer versions of TLAPM ignore?

In the current state of main branches, the Toolbox sends deprecated options that TLAPM ignores.

With the current state of this PR, deprecated options are removed, and TLAPM fails if they are used. So the Toolbox without https://github.com/tlaplus/tlaplus/pull/1085 will still fail with the new TLAPM.

ahelwer:
Excellent! I'll take a look. Do you think https://github.com/tlaplus/tlaplus/pull/1085 could be adapted to remove the unused deprecated CLI parameters?

src/tlapm_args.ml
```diff
@@ -134,140 +107,254 @@ let quote_if_needed s =
...
+let nonormal_t =
+  let doc = "Do not normalize obligations before printing." in
+  Arg.(value & flag & info ["nonormal"] ~doc)
+
+let debug_t =
+  let doc = "Enable/disable debugging flags." in
+  Arg.(value & opt_all string [] & info ["debug"] ~docv:"[-]flag" ~doc)
+
+let toolbox_t =
+  let doc = "Toolbox mode." in
```
ahelwer, line 189: The doc printer doesn't seem to handle `(some (pair int int))` and just prints out:
```
 --toolbox=VAL
           Toolbox mode.
```

src/tlapm_args.ml
```diff
@@ -134,140 +107,254 @@ let quote_if_needed s =
...
+let toolbox_vsn_t =
+  let doc = "Toolbox protocol version." in
+  Arg.(value & opt int 1 & info ["toolbox-vsn"] ~docv:"1|2" ~doc)
+
+let line_t =
+  let doc = "Line to prove." in
+  Arg.(value & opt (some int) None & info ["line"] ~docv:"LINENO" ~doc)
+
+let wait_t =
+  let doc = "Wait for $(docv) before printing obligations in progress." in
```
ahelwer, line 201: Add time units. Seconds?

src/tlapm_args.ml
```diff
@@ -134,140 +107,254 @@ let quote_if_needed s =
...
+  Arg.(value & opt int 3 & info ["wait"] ~docv:"TIME" ~doc)
+
+let stdin_t =
+  let doc = "Read the tla file from stdin instead of file system. \
```
ahelwer, line 205: The help output spacing is weird here:
```
           Read the tla file from stdin instead of file system.  Only  applies
           if single tla file is provided as input.
```

src/tlapm_args.ml
```diff
@@ -134,140 +107,254 @@ let quote_if_needed s =
...
+
+let safefp_t =
+  let doc = "Check tlapm, zenon, Isabelle versions for fingerprints." in
+  Arg.(value & flag & info ["safefp"] ~doc)
+
+let nofp_t =
+  let doc = "Do not use existing fingerprints, \
+             but overwrite any preexisting fingerprints associated \
+             with the current proof obligations, \
+             using the newly computed fingerprints and results" in
```
ahelwer, line 230: For consistency: missing period

src/tlapm_args.ml
```diff
@@ -134,140 +107,254 @@ let quote_if_needed s =
...
+  Arg.(value & flag & info ["nofp"] ~doc)
+
+let nofpl_t =
+  let doc = "Disable fingerprint use between given lines." in
```
ahelwer, line 234: Same issue with help printer just printing `VAL` instead of anything indicating input format

src/tlapm_args.ml
```diff
@@ -134,140 +107,254 @@ let quote_if_needed s =
...
              Alternatively, this directory \
              can be defined via the variable \
              `TLAPM_CACHE_DIR` of the runtime \
              environment. The command-line \
              parameter takes precedence over \
              the environment variable. If neither \
-             is specified, the default value is \".tlacache\".";
-  ]
+             is specified, the default value is \".tlacache\"." in
+  Arg.(value & opt (some dir) None & info ["cache-dir"] ~docv:"DIR" ~doc)
```
ahelwer, line 267: Could make `.tlacache` the default value of this parameter instead of spelling it out in the docs?

src/tlapm_args.ml
```diff
@@ -279,10 +366,41 @@ let init () =
...
+let cmd =
+  let info = Cmd.info "tlapm" in
+  Cmd.v info term
+
+let no_comma s =
+  match String.index s ',' with
+  | exception Not_found -> true
+  | _ -> false
+
+let rec preprocess = function
```
ahelwer, line 394: Add comment indicating this is a workaround to preserve backward compatibility for CLI parameter format.

ahelwer (changes requested):
Overall looks good. I would like to see some in-file tests for all the CLI options, their backwards-compatibility versions and not.

test/fast/fingerprint/load_v8_test.tla.disabled
```diff
@@ -27,7 +27,7 @@ THEOREM cantor ==
 ===============================================
 command: rm -rf load_v8_test.tlaps
 command: cp -r load_v8_test.tlaps.testbase load_v8_test.tlaps
-command: ${TLAPM} --toolbox 0 0 --isaprove ${FILE}
+command: ${TLAPM} --toolbox 0 0 ${FILE}
```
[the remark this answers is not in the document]
glondu: I suppose as well.

src/tlapm_args.ml
```diff
@@ -134,140 +107,254 @@ let quote_if_needed s =
...
-            "<directory> save auxiliary and \
-             temporary files under <directory>. \
+open Cmdliner
+
+let version_t =
+  let doc = "Show version number and exit." in
+  Arg.(value & flag & info ["version"] ~doc)
+
+let verbose_t =
+  let doc = "Show version number and exit." in
```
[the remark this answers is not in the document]
glondu: Right, I will fix that.

src/tlapm_args.ml
glondu: I will add a `~docv` for this option.

src/tlapm_args.ml
glondu: Yes, it indeed seems to be seconds.

src/tlapm_args.ml
glondu: Maybe this is due to Cmdliner justifying the text when displaying help in a terminal? "Weird" spacing disappears when using `--help=plain`.

src/tlapm_args.ml
glondu: OK.

src/tlapm_args.ml
glondu: OK.

src/tlapm_args.ml
glondu: OK.

src/tlapm_args.ml
glondu: I don't think this would be correct, since the command-line option overrides the environment variable `TLAPM_CACHE_DIR`, and `.tlacache` is used only when neither is given.

glondu:
> Excellent! I'll take a look. Do you think [tlaplus/tlaplus#1085](https://github.com/tlaplus/tlaplus/pull/1085) could be adapted to remove the unused deprecated CLI parameters?

Already done!

src/tlapm_args.ml
```diff
@@ -279,10 +366,44 @@ let init () =
...
+  Cmd.v info term
+
+let no_comma s =
+  match String.index s ',' with
+  | exception Not_found -> true
+  | _ -> false
+
+(* Preprocess command line before feeding it to Cmdliner, to preserve
+   backward compatibility. In particular, Cmdliner does not support
+   options with direct multiple arguments. *)
```
ahelwer, line 396: Could you add an example of the specific types of text that are being rewritten?

ahelwer:
Thanks! Did you see the comment about testing?

Issue #88, opened 2023-07-15
Nondeterministic tlapm install failure on ubuntu (building Pure failed)


ahelwer:
Using this release: https://github.com/tlaplus/tlapm/releases/tag/202210041448

Seen here: https://github.com/tlaplus/Examples/actions/runs/5531712640/jobs/10092766558

Error message:
```
2023-07-12 12:42:09 (159 MB/s) - ‘tlaps-1.5.0-x86_64-linux-gnu-inst.bin’ saved [151669984/151669984]

Installing the TLA+ Proof System

    Version: 1.5.0
Destination: /home/runner/work/Examples/Examples/deps/tlapm-install

Extracting ... done.
Compiling Isabelle theories.
Building Pure ...
Pure FAILED
(see also /home/runner/work/Examples/Examples/deps/tlapm-install/lib/tlaps/heaps/polyml-5.4.0_x86_64-linux/log/Pure)

   xml_of_proof (PBound i) = XML.Elem ((...), ...) |
      xml_of_proof (... ...) = XML.Elem (...) |
      xml_of_proof ... = ... ... |
      xml_of_proof ... = ... |
      ...
At (line 57 of "Tools/xml_syntax.ML")
structure XML_Syntax : XML_SYNTAX
val it = (): unit
poly: scanaddrs.cpp:107: PolyWord ScanAddress::ScanStackAddress(PolyWord, StackObject*, bool): Assertion `val.IsDataPtr()' failed.
make: *** [IsaMakefile:262: /home/runner/work/Examples/Examples/deps/tlapm-install/lib/tlaps/heaps/polyml-5.4.0_x86_64-linux/Pure] Error 134
/home/runner/work/Examples/Examples/deps/tlapm-install/lib/tlaps/Isabelle2011-1/lib/scripts/run-polyml: line 77:  1847 Aborted                 (core dumped) "$POLY" -q $ML_OPTIONS
FATAL ERROR: Could not compile Isabelle/Pure


    *** ABORT ***.
Error: Process completed with exit code 2.
```

ref https://github.com/tlaplus/Examples/issues/82

kape1395:
I think PR https://github.com/tlaplus/tlapm/pull/83 would solve this problem because that PR makes the Isabelle theories built at release time instead of installation time. Also, branch https://github.com/tlaplus/tlapm/tree/isabelle2020 switches the Isabelle to a much newer version.

ahelwer:
[...]

PR #179, opened 2024-11-27
Add retry loop for TLAPM build in CI

ahelwer:
Ref #88


kape1395:
What are the current reasons for the TLAPM build to fail?

ahelwer:
Don't know, that remains to be debugged at a later date.

PR #180, opened 2024-11-27
CI: Use macos-latest instead of versions 13 & 14

ahelwer:
@kape1395 is there a reason we run the CI on both macOS 13 and 14? Version 13 machines are very slow, taking about twice as long to complete as the rest. This PR just changes it to use macos-latest.


kape1395:
I don't know. Maybe @damiendoligez or @muenchnerkindl was running the older macos?

Issue #92, opened 2023-09-17
Introduce a `develop` branch?


kape1395:
Maybe it would be easier to integrate different changes if TLAPM had a `develop` branch along with the `master` (or `main`)?

In such a case, all the pull requests should be based on the `develop` branch and be merged first back to it. While the main branch is expected to be stable, the `develop` could have all the ongoing changes integrated and tested together. When the code quality in the `develop` branch is deemed ready for a release, it could be merged into the `master`, and a release would be produced based on it.

muenchnerkindl:
Sounds like a good suggestion to me!

kape1395:
Do you know how we can proceed with that? Probably that's:
 - The team has to decide on that. Or is the decision already made?
  - Someone from the maintainers should create the `develop` branch based on the `main` branch.
      - That should be a protected branch on Git Hub (commits can only be done via PRs to that branch).
      - If we encourage users to experiment with the development version, that branch could be made a default one. This is optional.
  - I would rebase my PRs to the `develop`.
  - Then, the approved PRs could be merged into that branch.

ahelwer:
Personal opinion, I think it would be better to keep main as the development branch. Upload the build of main to the Inria nightly build server, same as the TLA+ tools. Keep the commits flowing into main and don't worry too too much about the odd bug, use releases to mark the state of the branch as stable. See PR #96

ahelwer:
One obstacle to rolling releases is that the current head of the main branch does not do well on the tlaplus/examples corpus: https://github.com/tlaplus/tlapm/actions/runs/6680011626/job/18152686044

ahelwer:
@kape @lemmy could I be added as an owner of this repo to make it easier to deal with releases, CI secrets, etc.? Alternatively see if you can create a role that lets me manage those things without all the other privileges if you want.

lemmy:
Paging @damiendoligez and @muenchnerkindl.

muenchnerkindl:
I have no objection.

ahelwer:
Okay who has the permissions to make this change?

lemmy:
"ahelwer has been added as a collaborator with maintain permissions on the repository."

ahelwer:
This does give me the ability to draft new releases but I don't have access to CI settings. The "Code and automation" setting panel.

kape1395:
This is not relevant anymore. Let's use the main branch.

Issue #46, opened 2021-09-26
TLAPS does not install on macOS Big Sur (11.6)


lemmy:
```
“tlaps-1.4.5-i386-darwin-inst.bin” cannot be opened because the developer cannot be verified.
macOS cannot verify that this app is free from malware.
```

```
“tlaps-1.5.0-i386-darwin-inst.bin” cannot be opened because the developer cannot be verified.
macOS cannot verify that this app is free from malware.
```

```shell
armbook:Downloads markus$ ./tlaps-1.4.5-i386-darwin-inst.bin 
Killed: 9
armbook:Downloads markus$ ./tlaps-1.5.0-i386-darwin-inst.bin 
Killed: 9
```

lemmy:
Manual install on macOS Big Sur 11.6 M1:

```shell
$ brew install opam
$ brew install ocaml
$ opam init

$ git clone https://github.com/tlaplus/tlapm
$ cd tlapm

$ eval $(opam env)
$ ./configure
$ make
$ sudo make install
```

lemmy:
A thing I've been wondering about for a while: Why does the TLAPS installer build `polyml` instead of packaging a pre-build binary?

lemmy:
@will62794 The 1.4.5 installer didn't work for me on M1 MacBook Pro, macOS 11.6.  Did it work for you out of the box?

will62794:
@lemmy As far as I can recall, there were no major hiccups. I think I did the installation a few months ago (April 2021 perhaps?), using [these instructions](https://tla.msr-inria.inria.fr/tlaps/content/Download/Binaries/MacOS.html), and things went relatively smoothly. Looking at your errors above, though, I now seem to recall I may have had to work around some of the Mac permissions errors when trying to run the tlapm installer. See [here](https://zaiste.net/os/macos/howtos/resolve-macos-cannot-be-opened-because-the-developer-cannot-be-verified-error/) for a possible workaround.

lemmy:
TODO:

- [ ] Add suitable [entitlement.plist](https://github.com/tlaplus/tlaplustoolbox/org.lamport.tla.toolbox.product.product/entitlements.plist) for tlapm
- [ ] Copy "Apple_*" secrets from [github.com/tlaplus/tlaplus](https://github.com/tlaplus/tlaplus/settings/secrets/actions) to [org-level secrets](https://github.com/organizations/tlaplus/settings/secrets/actions)
[...]

PR #182, opened 2024-11-29
Upload rolling pre-releases

ahelwer:
Finally got around to this. This sets up rolling pre-release publishing similar to what exists for the TLA+ tools, where the head of the main branch is built and uploaded to a github release for ease of consumption.

This work requires some changes to the repository settings:
1. Create a `release` environment and constrain it to only be accessible from the `main` branch
2. Create a new pre-release titled (for example) `Version 1.6.0 rolling pre-release` or similar
3. Create a fine-grained personal access token with read/write access to releases in the tlaplus/tlapm repo
4. Store the access token in the `release` environment secrets as `TLAPM_RELEASES_AT`
5. Define the `ROLLING_PRERELEASE_VERSION` environment variable in the `release` environment to be `1.6.0-pre` or similar
6. Define the `ROLLING_PRERELEASE_GITHUB_NAME` environment variable in the `release` environment to be whatever you called the release in step (2)

If you give me more permissions on this repo I'll do all of that or whoever is reviewing this can do it themselves. I've tested this on my own fork of the TLAPM repo; you can see what the release with uploaded artifacts looks like here: https://github.com/ahelwer/tlapm/releases/tag/1.6.0-pre

This is a good first step to get this working and then later we can add the following features:
1. Update the tag associated with the release to track the head of the main branch
2. Do macOS code signing (ref #46)
3. Upload the releases to the Inria server as well

Ref #92

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
               run: |
                 sudo apt-get update
                 sudo apt-get install --yes time
-            - uses: ocaml/setup-ocaml@v2
+            - uses: ocaml/setup-ocaml@v3
               with:
                 ocaml-compiler: ${{ matrix.ocaml-compiler }}
             - uses: actions/cache@v4
@@ -53,7 +53,7 @@ jobs:
                 # Workaround for https://github.com/tlaplus/tlapm/issues/88
                 set +e
                 for ((attempt = 1; attempt <= 5; attempt++)); do
-                  rm -rf _build
+                  make clean
                   make
                   if [ $? -eq 0 ]; then
                     make release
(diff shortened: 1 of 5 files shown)


ahelwer:
@lemmy Stephan made me an admin of this repo, but it does not appear fine-grained personal access tokens are yet enabled for the tlaplus organization. Would you be able to enable them? Then I can create a PAT scoped only to modify TLAPM releases. Here are the instructions: https://docs.github.com/en/organizations/managing-programmatic-access-to-your-organization/setting-a-personal-access-token-policy-for-your-organization#restricting-access-by-personal-access-tokens

For token expiration policy I do think it's fine to leave it as unbounded (no expiration) but that's up to you. I believe you can also add the necessity of requesting approval before users can create PATs for repos under the tlaplus org; not sure how that flow works, guess we can try it.

ahelwer:
Actually I just learned about `GITHUB_TOKEN` so that won't be necessary. Might be useful to enable PATs in general though.

PR #186, opened 2024-12-24
Make the LSP independent of level assignment.

kape1395:
The fingerprint calculation reverted to is_const (as it is now done in the main branch) instead of get_level to make the fingerprint calculation independent of the level assignment. This is done to avoid having level assignments on the critical path in the LSP, as it is too slow. In my case, it took 30 seconds to calculate the levels, which blocks subsequent LSP commands.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,2 +1,2 @@
-version=0.26.2
+version=0.27.0
 profile=default
--- a/lsp/lib/docs/proof_step.ml
+++ b/lsp/lib/docs/proof_step.ml
@@ -370,7 +370,9 @@ let of_module (mule : Tlapm_lib.Module.T.mule) prev : t option =
            proof status between the modifications. *)
         let o =
           match o.fingerprint with
-          | None -> Tlapm_lib.Backend.Prep.prepare_obligation o
+          | None ->
+              (* `Tlapm_lib.Backend.Prep.prepare_obligation o` works too slow here. *)
+              Tlapm_lib.Backend.Fingerprints.write_fingerprint o
           | Some _ -> o
         in
         let o = Obl.of_parsed_obligation o Proof_status.Pending in
--- a/src/backend/fingerprints.ml
+++ b/src/backend/fingerprints.ml
@@ -446,9 +446,14 @@ and fp_sequent stack buf sq =
              spin stack cx;
              let v,r = Stack.pop stack in
              if !r then
+               (* Here `Expr.Levels.get_level e` was used instead of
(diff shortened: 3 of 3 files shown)


ahelwer:
The TLAPM parser has a level-checker?

kape1395:
It has, but I'm unsure how complete it is. See e.g. https://github.com/tlaplus/tlapm/blob/d8e2a2135063fbb0bd6a4fb3af7bba46951059c8/src/expr/e_levels.ml

kape1395:
Thanks for starting that discussion. I will try to join it, but my knowledge of that parser is limited.

damiendoligez (approved):
Let's merge this fix, I think it is the last thing holding up #148. We will revert this change when we figure out what's wrong with level inference.

kape1395:
Thanks @damiendoligez. I will spend another few evenings on the level checking performance, and if nothing good happens, I will merge this PR.

ahelwer:
All the transformations to the backend provers are defined over the existing parse tree format, is my current understanding.

lemmy:
Understood - there’s no issue with refactoring the parse tree format, but it will require refactoring some of the transformations.  To minimize the risk of regressions, such refactoring should be avoided unless justified.

kape1395:
> However, it makes sense to spin out the TLAPM parser into its own repo.

I don't think we need to keep the existing tlapm parser if the new one is integrated and works well.

Where can I put my wishes (would become requirements, if agreed) for the parser integration? Preliminary:
  - Should we add protobuf and/or grpc interface for the parser instead of parsing the XML? For performance reasons. LSP uses JSON-RPC, maybe we should follow that standard?
  - I would like to pass some checksums/modification times for the modules already parsed by the tlapm. This way, it could avoid repeatedly parsing them (the dependencies of a module currently edited). A stateful interaction is probably needed to handle that properly (that would imply grpc or similar, I guess).
  - I would like to have a way to parse a document not saved to the disk, e.g., from the stdin or grpc message payload. This is currently used by the tlapm lsp.

Also, it is unclear to me what happens to the expression levels when definitions in an expression are expanded. Don't we need to recalculate the levels? Does the SANY parser handle that?

ahelwer:
@lemmy good point. What do you think is a good term for the parameter of these not-higher-order-ops? I've seen "operator parameter parameter" a few times in the codebase which I don't really like.

lemmy:
> @lemmy good point. What do you think is a good term for the parameter of these not-higher-order-ops? I've seen "operator parameter parameter" a few times in the codebase which I don't really like.

I'm not aware a dedicated terminology.

Issue #151, opened 2024-09-03
tlapm ending abnormally with Invalid_argument("List.combine") in level checking


lemmy:
TLAPS has been terminating unexpectedly since the pre-release https://github.com/tlaplus/tlapm/releases/tag/202208050903. This issue appears to be linked to level checking.

```ocaml
$ tools/tlaps15pre/bin/tlapm --cleanfp BlockingQueueFair.tla
(* fingerprints file ".tlacache/FiniteSets.tlaps/fingerprints" removed *)
(* created new ".tlacache/FiniteSets.tlaps/FiniteSets.thy" *)
(* fingerprints written in ".tlacache/FiniteSets.tlaps/fingerprints" *)
File "/workspaces/BlockingQueue/tools/tlaps15pre/lib/tlaps/FiniteSets.tla", line 1, character 1 to line 23, character 77:
[INFO]: All 0 obligation proved.
(* fingerprints file ".tlacache/TLAPS.tlaps/fingerprints" removed *)
(* created new ".tlacache/TLAPS.tlaps/TLAPS.thy" *)
(* fingerprints written in ".tlacache/TLAPS.tlaps/fingerprints" *)
File "./TLAPS.tla", line 1, character 1 to line 311, character 77:
[INFO]: All 0 obligation proved.
(* fingerprints file ".tlacache/BlockingQueue.tlaps/fingerprints" removed *)
(* created new ".tlacache/BlockingQueue.tlaps/BlockingQueue.thy" *)
(* fingerprints written in ".tlacache/BlockingQueue.tlaps/fingerprints" *)
File "./BlockingQueue.tla", line 1, character 1 to line 152, character 77:
[INFO]: All 1 obligation proved.
 tlapm ending abnormally with Invalid_argument("List.combine")
Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
Called from Stdlib__list.combine in file "list.ml", line 305, characters 36-49
Called from E_levels.level_computation#expr.apply_operator.f in file "e_levels.ml", line 228, characters 28-55
Called from E_levels.level_computation#expr.apply_operator in file "e_levels.ml", line 234, characters 27-55
Called from E_levels.level_computation#expr in file "e_levels.ml", line 492, characters 23-40
Called from E_levels.level_computation#expr in file "e_levels.ml", line 620, characters 21-38
Called from E_action.lambdify in file "e_action.ml", line 2090, characters 12-39
[...]

PR #191, opened 2025-01-02
Fix instantiation of a module containing USE.

kape1395:
It was failing before with `tlapm ending abnormally with Failure("unknown bound variable")`.

This fixes a case mentioned in https://github.com/tlaplus/tlapm/issues/151#issuecomment-2365142957


muenchnerkindl:
I am afraid I know nothing about this part of the code, and I'd prefer if @damiendoligez or @johnyf could have a look.

Issue #192, opened 2025-01-04
LSP: USE should be used as separate steps


kape1395:
The `USE` statements produce proof obligations and thus should be shown as proof steps.

kape1395:
The same applies to the USE in a module, not just a proof step.
The former is represented by the `Mutate` variant of the AST.

PR #193, opened 2025-01-04
LSP: Create proof step markers for USE statements involving facts.

kape1395:
Fixes https://github.com/tlaplus/tlapm/issues/192

image


kape1395:
Some changes are caused by the `ocamlformat`, which started formatting the comments after its version was updated. I will create another PR to re-format other files under `./lsp`.

kape1395:
I skipped them when implementing the LSP mainly because these steps were not marked "green" in the toolbox. So, I assumed they produce no proof obligations. Now I understand they do.

ahelwer (approved):

PR #194, opened 2025-01-06
LSP: Use maximum obligation status instead of latest.

kape1395:
For a proof step, we will take the maximum of statuses returned by different provers for each obligation and the minimum across the obligations. Taking the latest sometimes failed because of race conditions from the prover, e.g., first reporting true from TLAPM and later receiving false from SMT, which resulted in false status.

It should be merged after #193.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,2 +1,2 @@
-version=0.26.2
+version=0.27.0
 profile=default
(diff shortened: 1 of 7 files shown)


kape1395:
@ahelwer, could you take a look at this?

ahelwer (approved):

ahelwer:
This looks like a good change!

PR #202, opened 2025-02-17
Fixed SMT soundness bug involving tuples and EXCEPT

### LINKED ISSUE #201: Soundness bug in TLAPS 1.6.0 pre-release
Following my encounter with the soundness bug reported in issue #200 and @muenchnerkindl’s recommendation to switch to the new SMT backend, I began adapting an earlier proof to be compatible with the current pre-release. During this process, I discovered a  soundness issue specific to the 1.6.0 pre-release. This issue is not present in earlier versions of TLAPS. The bug is present in the current build 41d03d4 and also 8138f4b (which I had previously built in September to test the LSP). 

```
--------------------------- MODULE SoundnessError ---------------------------
EXTENDS TLAPS, Naturals

THEOREM SoundnessError == FALSE
  <1> DEFINE f1 == [i \in Nat |-> i]
  <1> DEFINE f2 == [f1 EXCEPT ![0] = << 0 >>]
  <1>1. f1 = f2 OBVIOUS
  <1> QED BY <1>1

=============================================================================
```

The SMT backend incorrectly verifies `<1>1` even though it clearly is false.

rozlynd:
The bug originated from axiom "TupExcept" which specifies the replacement of a element in a tuple using EXCEPT.  For instance, the axiom could be used to generate the equality `[ <<a, b, c>> EXCEPT ![2] = d ] = <<a, d, c>>`.

The index of the element to change was not set correctly.  Also, the axiom is only put in the result if both a tuple and the EXCEPT operator appear in the obligation, making this bug harder to detect.

This closes #201

### DIFF
--- a/src/encode/n_axioms.ml
+++ b/src/encode/n_axioms.ml
@@ -2814,10 +2814,10 @@ let tupexcept_def ~disable_arithmetic n i =
       (ixi ~shift:1 n) %% []
     ; begin
       if disable_arithmetic then
-        apps (T.IntLit (i + 1)) [] %% []
+        apps (T.IntLit i) [] %% []
       else
         apps (T.Cast t_int)
-        [ apps (T.TIntLit (i + 1)) [] %% []
+        [ apps (T.TIntLit i) [] %% []
         ] %% []
       end
     ; Ix 1 %% []
@@ -2829,10 +2829,10 @@ let tupexcept_def ~disable_arithmetic n i =
         (ixi ~shift:1 n) %% []
       ; begin
         if disable_arithmetic then
-          apps (T.IntLit (i + 1)) [] %% []
+          apps (T.IntLit i) [] %% []
         else
           apps (T.Cast t_int)
-          [ apps (T.TIntLit (i + 1)) [] %% []
+          [ apps (T.TIntLit i) [] %% []
           ] %% []
         end
       ; Ix 1 %% []


ahelwer:
For DCO signoff you have to amend the commit with the -S flag. This is just a new thing since moving under the Linux Foundation.

rozlynd:
@ahelwer Okay, doing that now!

@muenchnerkindl This is indeed a valid example, I thought it would be less confusing to show that. Before the fix, you would find a '3' where there is a '2'.

ahelwer:
Maybe I'm doing something wrong but this doesn't seem to have fixed the issue:
```
> tlapm --version
e6d0831
> tlapm test/regression_tests/function_except_tuple_soundness_gh201.tla 
(* loading fingerprints in ".tlacache/function_except_tuple_soundness_gh201.tlaps/fingerprints" *)
(* created new ".tlacache/function_except_tuple_soundness_gh201.tlaps/function_except_tuple_soundness_gh201.thy" *)
(* fingerprints written in ".tlacache/function_except_tuple_soundness_gh201.tlaps/fingerprints" *)
File "./test/regression_tests/function_except_tuple_soundness_gh201.tla", line 1, character 1 to line 9, character 4:
[INFO]: All 3 obligations proved.
```
Running it on the example given in issue #201

muenchnerkindl:
Did you remove the fingerprints or try with option ``--nofp``?

ahelwer:
Yeah that was the issue, alas I realized right after posting lol. I'll add the files as regression tests.

PR #208, opened 2025-04-06
Support for proof step renames in the LSP server

kape1395:
Proof steps can be renamed by putting the cursor on a step or a reference to it and then pressing F2.

proof_step_rename

Some of the changes were needed to debug/understand the proof AST (e.g., `visitor_pp`). I think they will also be useful later.


ahelwer:
Could you provide a longer explanation of the various changes necessary for this functionality, the reasons things needed to be added, etc.? Focusing on the non-lsp part of the codebase.

kape1395:
@ahelwer, the changes to the non-LSP part are only these:
    - Export the `Builtin` package so that it is accessible to the users of the `Tlapm_lib` library (the LSP and others).
    - Introduce the `empty` function to create an empty context for traversing/visiting an expression.
    - All the rest are the functions to print internal representations of the data structures (the `pp_`  functions and `@@deriving show`).

ahelwer:
Okay, I'm signing off because those changes make sense and everything in the LSP folder is developed by you and can't introduce any soundness issues.

ahelwer (approved):

PR #209, opened 2025-04-10
Use stderr for informative reports.

kape1395:
This is to avoid interference with the LSP server, which uses the stdout for the protocol communication.

### DIFF
--- a/lsp/bin/tlapm_lsp.ml
+++ b/lsp/bin/tlapm_lsp.ml
@@ -14,6 +14,7 @@ let transport_of_args tr_stdio tr_socket =
 let run transport log_to log_io =
   Printexc.record_backtrace true;
   let main_fun (env : Eio_unix.Stdenv.base) =
+    Tlapm_lib.Util.redirect_stdout Format.err_formatter;
     let main_switch _sw =
       let stop_promise, stop_resolver = Eio.Std.Promise.create () in
       let handle_signal (_signum : int) =
--- a/src/backend/fpfile.ml
+++ b/src/backend/fpfile.ml
@@ -521,7 +521,7 @@ let fp_close_and_consolidate file oc =
   write_fp_table noc;
   close_out noc;
   Sys.rename tmpfile file;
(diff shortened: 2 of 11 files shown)


lemmy:
I think it would make sense to separate `stdout` and `stderr` even in the ordinary case.  While your terminal will display both streams, users can then redirect `stderr` using `2> tlaps.error`.

ahelwer (approved):

kape1395:
> Would it make more sense to put this as a mode of TLAPM?

Indeed, that's a good idea. This way I will be able to redirect all the possible output to the stderr, when executing it in the LSP server process. Nevertheless, separating the output to stdout and stderr for the ordinary execution should be done anyway. 

Also, there are several places where stdout is used directly, not via the Format or the Util module. That needs to be cleaned up as well.

Should I make these changes in this PR or a separate one?

ahelwer:
For things like:
```
Util.eprintf "(* fingerprints written in %S *)"
```
this seems like an ordinary diagnostic message and not an error, so it makes sense to output it on stdout instead of stderr. So in my mind making a LSP "mode" makes sense. The java tools have a `-tool` mode for example for running from the toolbox, and git has porcelain output mode. So there is precedence for this in other tools. However this ultimately does not matter very much and these changes are fine if they fix the LSP.

Might as well fix the standalone usages of stdout in their own PR that can be easily merged.

lemmy:
> The java tools have a `-tool` mode for example for running from the toolbox, and git has porcelain output mode. So there is precedence for this in other tools.

TLC’s `-tool` mode was a major design bug that is one reason why the Toolbox sucks. A much better approach would have been to use IPC like RMI (or a Java-independent alternative), instead of relying on inefficient and error-prone serializing and deserializing of TLC state.

kape1395:
@ahelwer, I have reverted the previous change and made it more in "tool mode". The change now looks cleaner.

damiendoligez (approved):
Very nice and clean, obviously correct.

PR #214, opened 2025-06-09
Encode TLAPM build dependencies in shell.nix file

ahelwer:
Got about as far as my nix knowledge allows (which is not too far!) in building tlapm from NixOS. @calvin-l have any idea how to make progress beyond this? When running `nix-shell shell.nix` and then `make`, it gets pretty far then:
```
make[2]: Leaving directory '/mnt/data/ahelwer/src/tlaplus/proofs/_build/default/deps/isabelle/Isabelle/src/TLA+'
cd Isabelle/ \
        && ./bin/isabelle build -o system_heaps -o document=false -b -v -d src/Pure Pure \
        && ./bin/isabelle build -o system_heaps -o document=false -b -c -v -d src/TLA+ TLA+ \
        && rm -rf ./heaps/polyml-*/log/*
make[1]: Leaving directory '/mnt/data/ahelwer/src/tlaplus/proofs/_build/default/deps/isabelle'
Could not start dynamically linked executable: /mnt/data/ahelwer/src/tlaplus/proofs/_build/default/deps/isabelle/Isabelle/contrib/jdk-21.0.3/x86_64-linux/bin/java
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
make[1]: *** [dune.mk:84: Isabelle/src/TLA+] Error 127
make: *** [Makefile:34: build] Error 1 
```
So it seems like it's invoking a general-release build of Java for the Isabelle build process.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,39 @@
+let pkgs = import <nixpkgs> {};
+in pkgs.mkShell {
+  packages = with pkgs; [
+    # Fundamental dependencies
+    bash
+    ocaml                       # OCaml compiler & interpreter
+    dune_3                      # OCaml build system
+    gnumake                     # Called by Dune to build non-OCaml code
+    ocamlPackages.findlib       # OCaml library manager
+
+    # CLI tools necessary to build & test project
+    gawk
+    gnupatch
+    time
+    zlib
+
+    # OCaml packages necessary to build & test project
+    ocamlPackages.camlzip         # Library for handling zip & gzip files
+    ocamlPackages.cmdliner        # Enables declarative CLI definitions
+    ocamlPackages.eio_main        # Parallel IO library
+    ocamlPackages.dune-build-info # Embed build information inside executables
+    ocamlPackages.dune-site       # Embed location information inside executables
+    ocamlPackages.lsp             # LSP protocol library for building LSP servers
+    ocamlPackages.odoc            # Documentation generator
+    ocamlPackages.ounit2          # Unit testing framework for OCaml
+    ocamlPackages.ppx_assert      # Macro library for assertions with useful errors
+    ocamlPackages.ppx_deriving    # Macro library for auto-displaying datatypes
+    ocamlPackages.ppx_inline_test # Macro library for writing in-line tests
+    ocamlPackages.re2             # Regular expression library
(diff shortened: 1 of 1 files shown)


Calvin-L:
Out-of-the-box, the Isabelle distribution is not friendly to Nix's hermetic ecosystem. To address the immediate issue: [Isabelle is already available as a Nix package with all the right patches](https://github.com/NixOS/nixpkgs/blob/fe51d34885f7b5e3e7b59572796e1bcb427eccb1/pkgs/by-name/is/isabelle/package.nix#L132), and you should pull that in instead of letting TLAPS download Isabelle binaries.

(That's also true for _all_ of TLAPS's dependencies.)

You might take a look at [how I build TLAPS using Nix](https://github.com/Calvin-L/calvins-nix-packages/blob/dc5c0712d8982aed3e905737bf40d38dc2c83f49/packages/tlaps.nix). That builder ignores TLAPS's makefiles entirely. Instead, it uses Isabelle from Nix and just invokes `dune` directly. (Well, it uses [a copy-pasted old expression for building Isabelle-2024](https://github.com/Calvin-L/calvins-nix-packages/blob/dc5c0712d8982aed3e905737bf40d38dc2c83f49/packages/isabelle-2024.nix), since Nix currently ships Isabelle-2025).

----

I have a lot more thoughts about this stuff. Here are a few (far from all):

The TLAPS build system (like every build system) needs to support multiple use cases, e.g.
 1. "Developer mode" where I don't need any complicated setup and the one-liner build command in the readme just works <-- most of us live in this universe
 2. "Package maintainer mode" where nothing gets downloaded and instead the package managers gives the build system paths to all its dependencies <-- nixpkgs maintainers, apt repo maintainers, yum repo maintainers, and so on live in this universe

One of Nix's strengths is that if you can make "package maintainer mode" work then you get "developer mode" almost for free.

TLAPS actually seems close to having support for both of those use cases. The makefiles are good for developers; they do a lot of extra work on your behalf. And you can always invoke `isabelle` + `dune` directly if you are a package maintainer. With Nix, maybe the right approach is to ignore the makefiles entirely, since their functionality overlaps with what Nix gives you?

PR #215, opened 2025-06-13
fix for C++11 string literal syntax change

damiendoligez:
My C++ compiler (latest MacOS) complains about having an identifier right after a string literal, without a space inbetween. Apparently this corresponds to a new syntax introduced in C++11.

The fix is to add a space, and it should be completely compatible with other C++ compilers.

The problem occurs in the LS4 source code, so the fix is in the patch we apply to LS4 before compiling.


muenchnerkindl (approved):
LGTM, thanks!

PR #216, opened 2025-06-16
Fix SubSeqTyping axiom for the empty case

rozlynd:
The corrected axiom lets you deduce `SubSeq(s, m, n) \in Seq(S)` when `n < m`. This is correct as the expression denotes the empty sequence in that case.


muenchnerkindl:
Thanks @rozlynd! 
@all: This is part of ongoing work and I'll make a few more changes, please ignore the PR for now.

PR #217, opened 2025-06-16
Isabelle2025

muenchnerkindl:
This PR upgrades the Isabelle backend to Isabelle2025 and also fixes all proofs in the standard library of theorems. This also required a subtle change to the SMT backend contributed by @rozlynd.

### DIFF
--- a/deps/isabelle/dune.mk
+++ b/deps/isabelle/dune.mk
@@ -4,17 +4,17 @@
 OS_TYPE=$(patsubst CYGWIN%,Cygwin,$(shell uname))
 HOST_CPU=$(shell uname -m)
 
-ISABELLE_VSN=Isabelle2024
+ISABELLE_VSN=Isabelle2025
 
 ifeq ($(OS_TYPE),Linux)
-	ISABELLE_SHA256=603aaaf8abea36597af3b0651d2c162a86c0a0dd4420766f47e5724039639267
+	ISABELLE_SHA256=3d1d66de371823fe31aa8ae66638f73575bac244f00b31aee1dcb62f38147c56
 	ISABELLE_ARCHIVE=$(ISABELLE_VSN)_linux.tar.gz
 	ISABELLE_ARCHIVE_TYPE=tgz
 	ISABELLE_ARCHIVE_DIR=$(ISABELLE_VSN)
 	FIND_EXEC=-executable
 endif
 ifeq ($(OS_TYPE),Darwin)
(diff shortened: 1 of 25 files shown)


kape1395:
Several proofs fail, at least on my PC. Not sure if that's hardware or OS. I'm using Linux.
  -  https://github.com/tlaplus/tlapm/blob/403f6318af37267bfcc9d4d9f473c8449b71bc4e/library/FunctionTheorems_proofs.tla#L926 - fails when checking from IDE, but passed when checked using `tlapm` directly (after cleaning cache).
  - These fail from CLI as well:
      - https://github.com/tlaplus/tlapm/blob/403f6318af37267bfcc9d4d9f473c8449b71bc4e/library/NaturalsInduction_proofs.tla#L192 
      - https://github.com/tlaplus/tlapm/blob/403f6318af37267bfcc9d4d9f473c8449b71bc4e/library/SequencesExtTheorems_proofs.tla#L311
      - https://github.com/tlaplus/tlapm/blob/403f6318af37267bfcc9d4d9f473c8449b71bc4e/library/SequencesExtTheorems_proofs.tla#L323
      -

ahelwer:
I actually don't think the library proofs are checked in the CI, so yeah we could do that. They also fail on the main branch build for me.

muenchnerkindl:
> They also fail on the main branch build for me.

Yeah, that's why I (tried to) fix them. ;-)

ahelwer:
That makes sense! Okay, I'll add the library proofs to the CI and comment out the ones that are failing. Then this PR can un-comment those proofs, and we will see whether they are fixed from the semi-objective perspective of the github CI.

ahelwer:
Okay if you rebase this PR on the latest changes in main, you can get the CI to check specific library proofs by deleting these lines:
https://github.com/tlaplus/tlapm/blob/867b94981ef9f523c25a7d3486eb1bd35b640171/.github/workflows/ci.yml#L68-L74

muenchnerkindl:
Thanks @ahelwer! So I commented out the line for FunctionTheorems_proofs.tla but I don't see the CI running. Do I have to start it manually for the PR?

muenchnerkindl:
OK, GitHub confirms that the proofs work on MacOS and fail on Linux. I even tried to decompose the failing proof, but still the same result. I'll install a local Linux box and continue testing there.

kape1395:
Two steps were failing on my PC in `SequencesExtTheorems_proofs.tla`. I made a PR #220 fixing them.

ahelwer:
The CI now takes 24 minutes instead of 18, which is not too bad. It's important for the library proofs to continue passing, I think? I will move the library checks later in the CI though so basic unit tests run first.

PR #220, opened 2025-06-21
Decompose proofs so that they pass on my PC.

kape1395:
This makes the library proofs passing on my PC.
Strangely, `<2>2` needed to be separated from `<2>1` to pass.


muenchnerkindl (approved):
Thank you very much, you saved me another roundtrip between Linux and Mac!

PR #222, opened 2025-06-22
Make LSP independent of level calculations (2nd attempt).

kape1395:
Added an option to ignore the level information while calculating a fingerprint.
By default, it is false; thus, the existing behaviour is retained.
The fingerprints with levels ignored are only used as hash functions to transfer proof state between the document versions in the LSP.

This time, the tests are passing.

### DIFF
--- a/lsp/lib/docs/proof_step.ml
+++ b/lsp/lib/docs/proof_step.ml
@@ -392,7 +392,12 @@ let of_module (mule : Tlapm_lib.Module.T.mule) prev : t option =
            proof status between the modifications. *)
         let o =
           match o.fingerprint with
-          | None -> Tlapm_lib.Backend.Prep.prepare_obligation o
+          | None ->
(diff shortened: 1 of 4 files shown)


ahelwer:
What tests were failing? Also what alternative approaches do you think there could be instead? For example, not using fingerprint files in the LSP

kape1395:
> What tests were failing?

Don't have the logs now. But 6 tests of `make test` were failing. Related to the levels. It looks like CI only runs tests if the PR is to the main branch.

> Also what alternative approaches do you think there could be instead? For example, not using fingerprint files in the LSP

The fingerprint files are not used. Just the fingerprint values. In principle, I have to have a way to understand that an obligation is the same on different versions of the document (between edits). Any hash function that would be agnostic to line numbers or particular formatting would do fine. The fingerprint is much like it, so I think it is not worth implementing that hash function aside, if it looks similar to the existing one.

damiendoligez (approved):
The code looks good and I find this a rather elegant way to solve the problem.

PR #224, opened 2025-06-23
CI: only run tests for OCaml 5 or higher

ahelwer:
@kape1395 thoughts?


kape1395:
I don't know. Debian stable (bookwork) still ships with 4.13. That was the argument to support OCaml version 4 last time we discussed this. If we support it, we should run the tests there as well.

ahelwer:
That makes sense. Should we make a 4.13 ocaml release as well then or does it not matter?

Issue #171, opened 2024-10-28
Functions.tla and SequencesExt.tla out-of-sync with CommunityModules


lemmy:
The [Functions.tla](https://github.com/tlaplus/tlapm/blob/main/library/Functions.tla) file in TLAPS is not aligned with the [Functions.tla](https://github.com/tlaplus/CommunityModules/blob/master/modules/Functions.tla) version from the CommunityModules repository. Similarly, the TLAPS version of [SequencesExt.tla](https://github.com/tlaplus/tlapm/blob/main/library/SequencesExt.tla) has diverged from the [SequencesExt.tla](https://github.com/tlaplus/CommunityModules/blob/master/modules/SequencesExt.tla) in CommunityModules. This discrepancy causes issues in VSCode when the LSP proof extension is enabled, as TLAPS overrides the CommunityModules versions, preventing TLC from accessing the latest definitions.

```shell
TLC2 Version 2.20 of Day Month 20?? (rev: cc65eef)
Running breadth-first search Model-Checking with fp 61 and seed -8162497252886980334 with 1 worker on 10 cores with 7282MB heap and 64MB offheap memory [pid: 46948] (Mac OS X 15.0.1 aarch64, Homebrew 11.0.25 x86_64, MSBDiskFPSet, DiskStateQueue).
Starting SANY...
Parsing file /Users/markus/src/TLA/_specs/MSFT/CCF/tla/consensus/MCabs.tla
Parsing file /Users/markus/src/TLA/_specs/MSFT/CCF/tla/consensus/abs.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/tlc-8661331616785400850/TLC.tla (jar:file:/Users/markus/.vscode/extensions/tlaplus.vscode-ide-2024.10.192237/tools/tla2tools.jar!/tla2sany/StandardModules/TLC.tla)
Parsing file /Users/markus/.opam/5.1.0/lib/tlapm/stdlib/SequencesExt.tla
Parsing file /private/var/folders/7d/4x6z2cc91jl588ysynlc1tfc0000gn/T/tlc-8661331616785400850/FiniteSetsExt.tla (jar:file:/Users/markus/.vscode/extensions/tlaplus.vscode-ide-2024.10.192237/tools/CommunityModules-deps.jar!/FiniteSetsExt.tla)
[...]

PR #239, opened 2025-09-05
Suffix module names with 'Fork' to avoid conflicts with newer versions of the same modules included in CommunityModules

lemmy:
Workaround for Github issue #171
https://github.com/tlaplus/tlapm/issues/171

[Bug]


ahelwer:
I don't think this is a good approach and seems downstream of the general lack of a well-thought-out versioning or packaging system for the community modules.

lemmy:
This is blocked because the proofs fail for the latest versions of `SequencesExt` and `Functions`.

PR #241, opened 2025-10-14
LSP: Proof decomposition code actions.

kape1395:
To try this out, build and install (`make all install`) TLAPM from this branch. The existing VSCode extension works without changes.

The majority of the decomposition suggestions are proposed for a QED step.

Please report your impressions on what could be improved.


lsp/lib/analysis/step_decompose/of_assm.ml
```diff
@@ -0,0 +1,397 @@
...
+  let title =
+    Fmt.str "⤮ Use (⇒): %a" (Debug.pp_expr_text (make_cx_hidden cx)) title_ex
+    |> limit_title
+  in
+  let ca =
+    ca_edits ~uri ~title ~edits:[ add_steps_rewrite; ps_proof_rewrite ]
+  in
+  [ ca ]
+
+(** For each assumption in the form of existential quantifier,
```
damiendoligez:
```suggestion
(** For each assumption in the form of universal quantifier,
```

lsp/lib/analysis/step_decompose/of_goal.ml
```diff
@@ -0,0 +1,352 @@
...
+           |> add_defs_from_pf ps_proof))
+  in
+  let ca =
+    ca_edits ~uri ~title:"⤮ Decompose goal (∧)"
+      ~edits:[ add_steps_rewrite; ps_proof_rewrite ]
+  in
+  [ ca ]
+
+(** Make decomposition code actions for a goal of the form of disjunction. In
+    general we leave one disjunct as a goal and assume all the rest are negated.
```
damiendoligez, line 192: You could also negate all disjuncts and leave FALSE as the goal, if you want to leave the user with a smaller set of choices.

damiendoligez (approved):
Looks good to me, only a couple of small remarks.

lsp/lib/analysis/step_decompose/of_assm.ml
kape1395: Thanks!

lsp/lib/analysis/step_decompose/of_goal.ml
kape1395: Interesting! @muenchnerkindl, what do you think about this change from the usage perspective?
The one implemented sounded natural to me, but proposing a bunch of alternatives is not always convenient. I can change that if we agree that proof by contradiction is more useful here.

muenchnerkindl: Negating all disjuncts is certainly more symmetric. It is not immediately clear which of the disjuncts should remain as the goal, and if I understand correctly, you leave the choice to the user, which requires an interaction that could be eliminated.

kape1395: I introduced the decomposition by negating all the disjuncts and proving false.
The old behaviour is still here and can be enabled by the configuration (disabled by default).

kape1395:
@muenchnerkindl, the build is now failing because of the missing `Functions.tla`? Maybe I should re-add that file for now, until a better resolution for the dependencies is found? Or should I merge this PR as is?

Issue #252, opened 2026-03-01
TLAPM library modules depend on community modules


ahelwer:
Recent changes have made it so that the TLAPM library modules require the community modules as a dependency (ex. FiniteSetTheorems.tla extending Functions.tla). This isn't great for usability for three reasons:
1. The community modules are not distributed with TLAPM
2. The dependency of the TLAPM library on the community modules is not documented anywhere
3. The usual community modules distribution method (a .jar file) is not accepted by TLAPM (although I implemented this feature in the [SANY work](https://github.com/ahelwer/tlapm/tree/sany-slice) which is not yet merged)

So I think there needs to be some thought put into how to resolve this. It also connects to a larger issue which is that we have allowed TLA+ standard library drift to occur. I suppose we currently have a hierarchy of "standard-ness" that goes:

1. TLA+ built-in operators (ENABLED, \lnot, =, /=, etc.)
2. The TLA+ standard modules as defined in Specifying Systems (Naturals, FiniteSets, Sequences, TLC, etc.)
3. Various extension modules distributed with TLC/SANY (TLCExt, Randomization, Json, etc.)
4. The TLAPM library modules (TLAPM, FiniteSetTheorems, BagsTheorems, etc.)
5. Apalache modules
6. The community modules

This seems like something the TLA+ Language Committee should be worrying about. Personally I favor a pretty strict approach where we should not allow easy mixing of levels in this hierarchy. For example, the extension modules in (3) probably should be hidden behind a feature flag in the Java tools, as otherwise they de-facto become part of (2) through use. It might even be too late for that, which would be unfortunate. I also don't think anything on this hierarchy should require anything ranked lower than it as a dependency, which is currently the case for the TLAPM library.

Ultimately this is design work that we should be thinking about very carefully. A language's libraries tend to be just as durable as its syntax, so changes and additions should go through the same standardization process as anything else.

muenchnerkindl:
Historically, the standard modules from Specifying Systems were distributed with TLC and baked into the PM (in the sense that as long as a module extends or instantiates one of these modules, the encoding recognizes these operators and handles them in an appropriate way). Then TLAPS needed a standard library of lemmas, and the opportunity was taken to define extra operators (the Functions module is a good example). In parallel, the Community modules started to collect operators that could be useful, and it was found that some of the library modules from TLAPS could be useful – but unsurprisingly, the two copies diverged. Apalache similarly built up its own modules.

[...]

Issue #213, opened 2025-05-27
Replace TLAPM parser with SANY


ahelwer:
As discussed in https://github.com/tlaplus/rfcs/issues/16, the TLAPM parser is to be replaced with SANY. This work has been approved for funding by the TLA⁺ Foundation. The reasons for this replacement are as follows:
1. Reduced maintenance burden, since only one fully-featured parser will need to be maintained
2. The TLAPM parser [lacks](https://github.com/tlaplus/rfcs/issues/16#issuecomment-2740711346) many semantic checks and does not have a fully-functional level checker
3. The TLAPM parser has [many bugs](https://github.com/tlaplus/tlapm/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22syntax%20parser%22) that need to be fixed (although most are minor)

One remaining question is how to handle the interface between TLAPM and SANY, as TLAPM is an OCaml program and SANY is a Java program. Three possibilities exist:
1. From TLAPM, call SANY's [XML Exporter](https://github.com/tlaplus/tlaplus/blob/96056d942b70656e7f88349e2f1fa38514c48783/tlatools/org.lamport.tlatools/src/tla2sany/xml/XMLExporter.java) program to parse the TLA⁺ input in a separate process, then [consume the output XML from OCaml](https://github.com/ahelwer/tlapm/commit/003f409442bc1fdf6622513ff63ad1fe16e78c97) & translate it to TLAPM's existing abstract syntax tree (AST) format; to reduce startup time, SANY will be run as a [GraalVM Native Image](https://github.com/tlaplus/tlaplus/pull/1177) so it does not need to load the JVM.
2. Compile SANY to a GraalVM Native Image, exposing a [C API](https://www.graalvm.org/latest/reference-manual/native-image/native-code-interoperability/C-API/) so SANY can be called directly from C code; then, call this C API [from OCaml](https://ocaml.org/docs/calling-c-libraries) code in TLAPM.
3. Call SANY from TLAPM OCaml code using some other OCaml/Java interface method; there are a few ([1](https://github.com/xavierleroy/camljava), [2](https://github.com/xclerc/ocamljava), [3](https://github.com/Julow/ocaml-java)) of these projects floating around but nothing that seems fully-featured & maintained, thus this option is not seriously considered.

[...]

PR #255, opened 2026-03-28
Add SANY parser backend

ahelwer:
The big files to look at here are:
- `src/sany/xml.ml`: this shells out to SANY and parses its output into a collection of typed datastructures that reflect the XML output format; the XML output format documentation was added in https://github.com/tlaplus/tlaplus/pull/1346)
- `src/sany/sany.ml`: this takes the datastructures generated by `src/sany/xml.ml` and converts it into TLAPM's existing parse tree

The rest is test code, and modifications to `Tlapm_lib` to add entrypoints to the SANY functionality hidden behind a feature flag.

The current status of the work is as follows:
- All files in the tlaplus/examples repo (and all .tla files in this repo) are parsed with SANY and have the XML output successfully run through the conversion process. Success here means the conversion process doesn't produce exceptions, although recursive operators and anonymous subexpressions are not supported. This test is currently only hardcoded to work on my personal computer since the CI of this repository is broken anyway due to #252.
- A selection of basic proofs are fully checked after being sent through the SANY parse pipeline. Extending this to check every proof in the tlaplus/examples repo likely would not take much more work.
- Functionality is exposed to compare the S-expressions of the parse trees produced by TLAPM and SANY for differences. There are a couple unbridgeable gaps (like SANY exports `VARIABLES x, y` as `VARIABLE x` then `VARIABLE y`) but after accounting for these comparing the S-expression output is a very nice way to spot differences. S-expression export has not yet been added on the SANY side.
- There are quite a few nitpick bugs that need to be fixed on the SANY XML Exporter side. I will be documenting those in the near future; a good central issue for now is https://github.com/tlaplus/tlaplus/issues/1313 (this is one of those things where the more you look the more you find so I tried to focus on getting this to shippable state instead of getting bogged down in making it perfect on both sides).

Ref #213


ahelwer:
Superseded by #275

PR #256, opened 2026-03-30
Added community modules as a git submodule

ahelwer:
Should fix CI and unit test failures

Ref #252

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,8 @@ jobs:
         steps:
             - name: Clone repo
               uses: actions/checkout@v4
+              with:
+                submodules: recursive
             - name: Install deps (ubuntu)
               if: matrix.os == 'ubuntu-latest'
               run: |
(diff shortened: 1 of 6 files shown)


library/CommunityModules
lemmy, line 1: I recommend following the main branch instead of pinning it to a specific commit.

ahelwer: I am not entirely sure that is possible, but I tried nonetheless (see .gitmodules file)

lemmy: If Git itself doesn’t support this, one option would be to have tlapm’s GitHub automation download and extract CommunityModules.zip as part of the build process.

This wouldn’t solve the local development setup, but in practice that already requires fetching dependencies (e.g., via opam), so this wouldn’t be introducing a fundamentally new requirement.

More importantly, it would align tlapm with how other projects (tlaplus/tlapus and tlaplus/vscode-tlaplus) consume tlaplus/CommunityModules.

ahelwer:
Closing in favor of an approach where the build system runs a wget on the latest CommunityModules release

Issue #136, opened 2024-06-06
Verify proofs in examples/ directory as part of PR workflow


lemmy:
The prover's build checks a set of tests, and the PR verifies proofs from tlaplus/examples.  However, the proofs in `examples/` are not verified.

kape1395:
I thought we would move these examples to the https://github.com/tlaplus/Examples/ repo. If not, I can add them to the test cases.

lemmy:
I don't have an opinion, except that some examples are perhaps too basic for the example repo?!

muenchnerkindl:
I suggest we do this on a case-by-case basis. In particular, I think that Allocator, Euclid, Peterson, and perhaps GraphTheorem could go to the GitHub Examples repo. Bakery, Byzantine and standard Paxos, EWD 840, and LamportMutex are in there already, but we should check if there is anything in tlapm/examples that is worth merging into the existing modules. The other examples look too basic to me, but should probably be added to the tests if they aren't there already.

If you agree, I can take care of that reshuffle.

kape1395:
Sounds like a good way to go.

lemmy:
`GraphTheorem` could be aligned with a moved next to https://github.com/tlaplus/CommunityModules/blob/master/modules/Graphs.tla, like `FiniteSetExtTheorem` is next to `FiniteSetExt` in the CommunityModules?

ahelwer:
Also noting that some of the specs in the examples dir don't actually parse (as decided by SANY):

paxos/Paxos.tla:
>line 406, col 31 to line 406, col 31 of module Paxos
Multiply-defined symbol 'v': this definition or declaration conflicts
with the one at line 371, col 18 to line 371, col 18 of module Paxos.
line 408, col 53 to line 408, col 53 of module Paxos
Multiply-defined symbol 'v': this definition or declaration conflicts
with the one at line 371, col 18 to line 371, col 18 of module Paxos.

ByzPaxos/BPConProof.tla:
>line 1481, col 23 to line 1481, col 23 of module BPConProof
Multiply-defined symbol 'r': this definition or declaration conflicts
with the one at line 1470, col 35 to line 1470, col 35 of module BPConProof.

GraphTheorem.tla
>line 140, col 20 to line 140, col 20 of module GraphTheorem
Multiply-defined symbol 'n': this definition or declaration conflicts
with the one at line 138, col 22 to line 138, col 22 of module GraphTheorem.

PR #257, opened 2026-04-06
Check local example proofs

younes-io:
#136

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,9 @@ jobs:
               run: |
                 find ./library -type f -iname "*_proofs.tla" -print0 |
                 xargs --null --max-args=1 ./_build/tlapm/bin/tlapm --cleanfp --stretch 5
+            - name: Check local example proofs
+              run: |
+                bash ./.github/scripts/check_local_examples.sh "./_build/tlapm/bin/tlapm"
             - name: Clone tlaplus/examples
               uses: actions/checkout@v4
               with:
@@ -108,4 +111,3 @@ jobs:
                   --tlapm_path "$DEPS_DIR/tlapm-install"        \
                   --examples_root "$EXAMPLES_DIR"               \
                   --skip "${SKIP[@]}"
-
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,10 +83,10 @@ jobs:
               run: |
                 # Skip the longest-running examples so PR CI stays under five minutes.
                 find ./examples -type f -name "*.tla"          \
-                ! -path "./examples/ByzPaxos/BPConProof.tla"  \
-                ! -path "./examples/ByzPaxos/VoteProof.tla"   \
+                ! -path "./examples/ByzPaxos/BPConProof.tla"   \
+                ! -path "./examples/ByzPaxos/VoteProof.tla"    \
                 -print0 | xargs --verbose --null --max-args=1  \
-                sh -c 'time ./_build/tlapm/bin/tlapm --cleanfp --stretch 5 "$1"' sh
+                /usr/bin/time ./_build/tlapm/bin/tlapm --cleanfp --stretch 5
             - name: Clone tlaplus/examples
               uses: actions/checkout@v6
               with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -80,6 +80,9 @@ jobs:
               run: |
                 find ./library -type f -iname "*_proofs.tla" -print0 |
                 xargs --null --max-args=1 ./_build/tlapm/bin/tlapm --cleanfp --stretch 5
+            - name: Check local example proofs
+              run: |
+                bash ./.github/scripts/check_local_examples.sh "./_build/tlapm/bin/tlapm"
```
ahelwer: Why can this not be done as a `find | xargs` one-liner like the library proofs check above it?

ahelwer:
Hi @younes-io are you interested in pushing this PR toward completion?

.github/workflows/ci.yml
```diff
@@ -79,6 +79,16 @@ jobs:
                 find ./library -type f -iname "*_proofs.tla"  \
                 -print0 | xargs --null --max-args=1           \
                 ./_build/tlapm/bin/tlapm --cleanfp --stretch 5
+            - name: Check local example proofs
+              run: |
+                # Skip examples with known parser/proof failures; see #136.
+                find ./examples -type f -name "*.tla"          \
+                ! -path "./examples/paxos/Paxos.tla"           \
```
ahelwer: These were fixed in https://github.com/tlaplus/tlapm/pull/251

younes-io: Thanks, addressed. `VoteProof.tla` is excluded because it currently fails locally with 2 unproved obligations.

ahelwer:
@younes-io thanks! These are working well; could you also wrap the TLAPM invocation within a `time` command so we can see how long each proof takes to check? Currently the new CI step takes 20 minutes which is a lot of time to add to the CI. Are there specific example proofs that take a lot of time that we could skip? Ideally this step would complete in less than five minutes.

muenchnerkindl:
@younes-io I am assuming this refers to examples/ByzPaxos/VoteProof.tla? I pushed a fix for these two steps to main by making the reasoning a bit more explicit, which hopefully makes the proof less brittle. Note, however, that some of the steps in the liveness proof are still omitted.

Also, it seems that you exclude examples/paxos/Paxos.tla, which works fine on my side. Does proof checking take too long for the GitHub runners?

Thanks!

ahelwer:
@younes-io can you remove the longest-running proofs so this step runs in under five minutes?

.github/workflows/ci.yml
```diff
@@ -79,6 +79,14 @@ jobs:
...
                 -print0 | xargs --null --max-args=1           \
                 ./_build/tlapm/bin/tlapm --cleanfp --stretch 5
+            - name: Check local example proofs
+              run: |
+                # Skip the longest-running examples so PR CI stays under five minutes.
+                find ./examples -type f -name "*.tla"          \
+                ! -path "./examples/ByzPaxos/BPConProof.tla"  \
+                ! -path "./examples/ByzPaxos/VoteProof.tla"   \
+                -print0 | xargs --verbose --null --max-args=1  \
+                sh -c 'time ./_build/tlapm/bin/tlapm --cleanfp --stretch 5 "$1"' sh
```
ahelwer: This shouldn't need to be run in a subshell

.github/workflows/ci.yml
```diff
@@ -79,6 +79,14 @@ jobs:
                 find ./library -type f -iname "*_proofs.tla"  \
                 -print0 | xargs --null --max-args=1           \
                 ./_build/tlapm/bin/tlapm --cleanfp --stretch 5
+            - name: Check local example proofs
+              run: |
+                # Skip the longest-running examples so PR CI stays under five minutes.
+                find ./examples -type f -name "*.tla"          \
```
ahelwer, line 85: nit: align backslashes

ahelwer (changes requested):

.github/workflows/ci.yml
younes-io: Fixed in 1f67a99: xargs now invokes /usr/bin/time directly.

.github/workflows/ci.yml
younes-io: Fixed in 1f67a99: aligned the continuation backslashes.

lemmy (approved):
The execution time is approximately 4.5 minutes on Linux and 5.5 minutes on a macOS runner. This is close enough to the requested five minutes.

PR #260, opened 2026-04-14
fix issue 212

muenchnerkindl:
Explicitly adding a step proved by OBVIOUS to a later proof appears to be necessary for Z3 on Linux. Sigh.


ahelwer:
You can also remove the line skipping the proof check in the CI!

Issue #265, opened 2026-05-04
SMT SetOf_n extensionality axiom uses wrong arity for n-ary comprehensions


ylht:
## Summary

The SMT encoding for `SetOf_n` extensionality introduces an incorrectly typed `AssertIsSetOf n` axiom when `n >= 2`.

`SetOf_n` is declared as taking `n` domain arguments plus one predicate/function argument, but `assert_issetof n` currently quantifies only one domain argument and gives the predicate type `idv -> idv`. This works accidentally for `n = 1`, but it is wrong for set comprehensions with multiple bound variables.

## Minimal reproduction

```tla
---- MODULE setof_nary_smt_test ----

EXTENDS TLAPS, Naturals

THEOREM SetOfTwoBoundVariablesUsesBinaryPredicate ==
  {x + y : x \in 1..1, y \in 1..1} = {2}
BY SMT

====
```

Run without fingerprints:

```sh
tlapm --cleanfp --nofp --cache-dir /tmp/tlapm-setof-nary-cache test/fast/regression/setof_nary_smt_test.tla
```

## Actual result

On `471b481` with fingerprints disabled:

```text
[ERROR]: 1/2 obligations failed.
```

## Expected result

The theorem should prove:

```text
[INFO]: All 2 obligations proved.
```

## Root cause

`src/encode/n_axioms.ml` defines `assert_issetof n` as if every `SetOf_n` had only one domain argument:

```ocaml
seq [ "P" ] [ Ty1 ([ t_idv ], t_idv) ]
...
apps (T.SetOf n) [ Ix 1; Ix 2 ]
```

But `src/encode/n_data.ml` declares `SetOf_n` as:

```ocaml
List.init n (fun _ -> t_cst t_idv)
@ [ Ty1 (List.init n (fun _ -> t_idv), t_idv) ]
```

So the axiom should quantify `n` domain arguments and call `SetOf_n(a1, ..., an, P)`.

The broken axiom appears to have been introduced by `8450a8c` (`Compare all pairs of refinement sets (heuristics)`), which added `AssertIsSetOf n` for `SetOf n when ext`.

muenchnerkindl:
Thank you for your report – however, the issue is well known and much broader than that: TLAPS does not currently support constructs binding more than one variable, see also https://proofs.tlapl.us/doc/web/content/Documentation/Unsupported_features.html.

PR #266, opened 2026-05-04
Fix SMT SetOf n-ary extensionality axiom

ylht:
## Summary

Fix the SMT extensionality helper axiom for `SetOf_n` so n-ary set comprehensions pass all `n` bound-domain arguments to the SMT `SetOf_n` application.

Refs tlaplus/tlapm#265.

## Scope

This PR is intentionally limited to an SMT backend arity fix. It does not add general TLAPS support for TLA+ constructs that bind more than one variable, and it does not address support in the Zenon or Isabelle backends.

The broader multi-variable binding limitation is discussed in tlaplus/tlapm#265 and in the TLAPS unsupported features documentation.

## Root Cause

`SetOf_n` is declared as taking `n` domain arguments plus one predicate/function argument:

```ocaml
SetOf_n(a1, ..., an, P)
```

But `assert_issetof n` encoded the assertion as if every `SetOf_n` were unary:

```ocaml
SetOf_n(a, P)
```

This is accidentally correct for `n = 1`, but wrong for `n >= 2`.

## Changes

- Give `P` the type `idv^n -> idv` in `assert_issetof n`.
- Quantify `n` bound-domain arguments.
- Build `SetOf_n(a1, ..., an, P)` in the trigger and assertion.
- Add a minimal SMT regression test using a two-variable set comprehension.

## Validation

New regression test on the fixed binary:

```sh
tlapm --cleanfp --nofp test/fast/regression/setof_nary_smt_test.tla
```

Result:

```text
[INFO]: All 2 obligations proved.
```

The same test fails on `471b481` with fingerprints disabled:

```text
[ERROR]: 1/2 obligations failed.
```

Build:

```sh
DUNE_PROFILE=release OCAMLPARAM='_,O3=1,unbox-closures=1,rounds=2' opam exec -- dune build src/tlapm.exe
```


muenchnerkindl:
As noted in the related issue, TLAPS not supporting constructs binding more than one variable applies in a much broader sense than just for the SMT backend. While your fix may go some way for introducing support for the SMT backend specifically, supporting such constructs needs more pervasive changes to tlapm and the different backends. Note that
```
THEOREM {x+y : x \in {1}, y \in {1}} = 2
OBVIOUS
```
outputs error message concerning SMT, Zenon, and Isabelle.

ylht:
Thanks, that makes sense. I agree this PR does not implement general support for multi-variable binders across TLAPS and all backends.

I will update the PR description to make the scope explicit: this is only an SMT backend arity fix for the existing SetOf_n encoding, not a full fix for #265. I will also change "Closes #265" to "Refs #265".

If you prefer not to accept partial SMT-only support for this unsupported construct, I can close the PR.

PR #268, opened 2026-05-11
Proof decomposition by template.

kape1395:
Extensible proof decomposition rules.


kape1395:
@uguryavuz, that's the branch I was working on. The match function is implemented only partially, for a PoC. Tried to rewrite it using existing visitors, but probably will stick with explicit traversal using a recursive function.

PR #273, opened 2026-05-17
CI: check example specs using find | jq | xargs

ahelwer:
Ref https://github.com/tlaplus/examples/issues/171

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,12 +22,10 @@ jobs:
               - '5.1.0'
         env:
           EXAMPLES_DIR: "tlaplus-examples"
-          SCRIPT_DIR: "tlaplus-examples/.github/scripts"
           DEPS_DIR: "tlaplus-examples/deps"
-          DIST_DIR: "tlatools/org.lamport.tlatools/dist"
         steps:
             - name: Clone repo
-              uses: actions/checkout@v4
+              uses: actions/checkout@v6
             - name: Install deps (ubuntu)
               if: matrix.os == 'ubuntu-latest'
               run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -82,31 +80,31 @@ jobs:
...
+                find "$EXAMPLES_DIR/specifications" -iname "manifest.json" -print0 \
+                | xargs --null --no-run-if-empty \
+                jq --join-output '
+                  .modules
+                  | map(select(has("proof")))
+                  # Failing on Linux
+                  | map(select(.path != "specifications/LoopInvariance/SumSequence.tla"))
+                  # Take too long to check in the CI
+                  | map(select(.path != "specifications/LoopInvariance/Quicksort.tla"))
+                  | map(select(.path != "specifications/bcastByz/bcastByz.tla"))
```
lemmy: Consider moving this information into the manifests for the individual spec directories. This would likely require a schema change, but since the manifests are the link between the TLA+ specs and CI, it seems preferable to keep this configuration in the manifest rather than in the CI setup itself.

ahelwer: We kind of have this info in the manifest with the proof runtime, which is probably how I would choose to implement. Note that the list of specs skipped here is longer than in the examples CI, since at least when doing development on this repo I enjoy a quicker CI than with the examples. So in some sense this set of skipped specs is more a function of the CI than the specs themselves. But yeah I guess filtering based on runtime would be the nice way to do it.

lemmy: The runtime field needs a reference system specification in order to be meaningful. Using the current generation of GitHub CI runners as that reference is problematic, because GitHub will eventually retire them, making the reported numbers impossible to reproduce.

I suggest adding a single boolean field to the manifest, such as `check_as_part_of_ci`, and reserving runtime for a future use case where it can be defined against a stable, reproducible reference environment.

Issue #277, opened 2026-06-07
`BY I!Thm` is opaque to backends when the source module contains `ASSUME`


tangruize:
## Summary

If module A has `ASSUME` declarations and a proven theorem `Thm`, then in module B, `I == INSTANCE A` followed by `BY I!Thm` silently fails, even when B declares the same assumptions.

The issue does not occur when (1) using `EXTENDS` instead of `INSTANCE`, or (2) the source module has no `ASSUME`.

### A.tla

```tla
---- MODULE A ----
EXTENDS Naturals, TLAPS

CONSTANT N
ASSUME NA == N > 0

THEOREM Thm == N > 0
  BY NA
====
```

### B.tla

```tla
---- MODULE B ----
EXTENDS Naturals, TLAPS

CONSTANT N
ASSUME NA == N > 0

I == INSTANCE A WITH N <- N

THEOREM Test == N > 0
  BY I!Thm
====
```

### Commands

```bash
$ ~/.tlapm/bin/tlapm  A.tla
(* created new ".tlacache/A.tlaps/A.thy" *)
(* fingerprints written in ".tlacache/A.tlaps/fingerprints" *)
File "./A.tla", line 1, character 1 to line 11, character 4:
[INFO]: All 2 obligations proved.

$ ~/.tlapm/bin/tlapm  B.tla
Zenon error: exhausted search space without finding a proof
(* created new ".tlacache/B.tlaps/B.thy" *)
(* fingerprints written in ".tlacache/B.tlaps/fingerprints" *)
File "./B.tla", line 11, characters 3-4:
[ERROR]: Could not prove or check:
           ASSUME NEW CONSTANT N,
                  NA == N > 0,
                  I!Thm
           PROVE  N > 0
File "./B.tla", line 1, character 1 to line 13, character 4:
[ERROR]: 1/2 obligations failed.
There were backend errors processing module `"B"`.
 tlapm ending abnormally with Failure("backend errors: there are unproved obligations")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Tlapm_lib.process_module in file "src/tlapm_lib.ml", line 448, characters 12-77
Called from Tlapm_lib.main.f in file "src/tlapm_lib.ml", line 577, characters 23-43
Called from Stdlib__List.fold_left in file "list.ml", line 123, characters 24-34
Called from Tlapm_lib.main in file "src/tlapm_lib.ml", line 580, characters 13-40
[...]

PR #279, opened 2026-06-18
doc: add practical hint on using theorems from instantiated modules

lemmy:
Document why BY I!Thm silently fails when an instantiated module has ASSUME declarations: the instance carries the (instantiated) assumptions as hypotheses, so the assumption must be discharged explicitly. Also cover the case where the assumption mentions an EXTENDS-imported operator, where a bridge theorem equating the I!-prefixed operator with its local counterpart is required. Based on https://github.com/tlaplus/tlapm/issues/277.

### DIFF
--- a/doc/web/content/Documentation/Tutorial/Practical_hints.html
+++ b/doc/web/content/Documentation/Tutorial/Practical_hints.html
@@ -195,6 +195,51 @@
   <p>The SMT backend should not require similar help for reasoning about
     records.</p>
 
+  <h3>Using theorems from instantiated modules</h3>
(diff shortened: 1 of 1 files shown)


muenchnerkindl:
Thanks, you were faster than me in figuring out how the documentation works! Is OCaml 4.14.1 really still supported by tlapm or should it be retired?

lemmy:
The build failure was a fluke.

lemmy:
@muenchnerkindl Is this good to go?

muenchnerkindl:
LGTM, thanks!

PR #282, opened 2026-07-13
Fix crash when a `DEF` clause cites a non-definition

### LINKED ISSUE #280: `tlapm` crashes with `Failure("Proof.Gen.set_defn/doit")` when a `DEF` clause cites a non-definition
### Description

When a `DEF` clause (`USE DEF X`, `HIDE DEF X`, or `BY ... DEF X`) cites a symbol that is in scope but is not a definition — a `CONSTANT`, a `VARIABLE`, or a `NEW`-declared constant/operator — `tlapm` crashes the whole run with `Failure("Proof.Gen.set_defn/doit")` ("Oops, this seems to be a bug in TLAPM").

Under `--verbose`, tlapm prints a correct, located diagnostic for this case — `Indicated point is not a definition, but NEW CONSTANT C!` — but then `failwith`s and aborts instead of reporting it as a normal error.

### Steps to reproduce

`Min.tla`:

```tla
---- MODULE Min ----
CONSTANT C
THEOREM TRUE BY DEF C
====
```

```
$ tlapm --verbose Min.tla
File "./Min.tla", line 3, character 21:
Error: Indicated point is not a definition, but NEW CONSTANT C!
 ...
 tlapm ending abnormally with Failure("Proof.Gen.set_defn/doit")
```

The same crash occurs for any non-definition, e.g. `USE DEF v` for a `VARIABLE v`, or `BY DEF xx` for a `NEW xx`.

### Expected behavior

Report "`C` is not a definition" as a normal located error and exit cleanly.

### Root cause

The crash is in `set_defn` in `src/proof/p_gen.ml`:

https://github.com/tlaplus/tlapm/blob/80172c61842c8dd15524c6f01a70ba91029802f5/src/proof/p_gen.ml#L50-L74

`set_defn vis l df` sets the visibility of the definition referenced by `df`. It is called from `gen_step` while folding over the `defs` of a `USE`/`HIDE` step; a `BY ... DEF` clause lowers to the same fold:

https://github.com/tlaplus/tlapm/blob/80172c61842c8dd15524c6f01a70ba91029802f5/src/proof/p_gen.ml#L314-L323

`df` is elaborated to a De Bruijn index, and `set_defn` walks to that context slot expecting a `Defn`. For a `CONSTANT`/`VARIABLE`/`NEW` the slot is a `Fresh`/`Fact` rather than a `Defn`, so the `Some (e, _)` branch prints the `~at:df` message under `--verbose` and then runs `failwith "Proof.Gen.set_defn/doit"`.

### Suggested fix

Turn that branch into a normal located error rather than a fatal `failwith`: emit the existing `~at:df` "not a definition" message unconditionally and report it through `Errors`, as the sibling empty-context branch already does with `Errors.bug ~at:df`. A non-definition `DEF` reference could also be rejected earlier during elaboration, where unknown names are already rejected with a locus.

### Environment

- `tlapm`: reproduced on commits 80172c6 and e2f60b8.

tangruize:
## Description

Fixes #280.

### Root cause

During anonymization (`src/proof/p_anon.ml`) a `DEF`-cited name that is in scope resolves to an `Ix` and was passed on unchecked as a `Dx`. When it names a declaration rather than a definition, `Proof.Gen.set_defn` later walked to a `Fresh`/`Fact` context slot expecting a `Defn` and aborted with `failwith`.

### Fix

Reject the reference already in `P_anon`, where the resolved index is available: if it does not point at a `Defn`, emit a normal located error (`"C" is not a definition and cannot be used in a DEF clause.`) and drop the reference, mirroring the existing handling of unexpandable identifiers. The run now reports the mistake and continues cleanly instead of crashing.

### Tests

Adds `test/bugs/def_nondef_test.tla` covering all three `DEF`-clause forms (`BY DEF`, `USE DEF`, `HIDE DEF`); it fails on the previous `set_defn` crash and passes with the fix.

### DIFF
--- a/src/proof/p_anon.ml
+++ b/src/proof/p_anon.ml
@@ -221,7 +221,21 @@ class anon = object (self : 'self)
                 let opaque = Opaque v @@ dw in
                 let opaque = self#expr scx opaque in
                 match opaque.core with
-                | Ix n -> [Dx n @@ dw]
+                | Ix n ->
+                    (* The name is in scope, but a DEF clause may only cite a
+                     * definition. If it resolves to a declaration
+                     * (CONSTANT/VARIABLE/NEW), reject it here with a located
+                     * error instead of crashing later in `Proof.Gen.set_defn`
+                     * (see issue #280). *)
+                    begin match Deque.nth ~backwards:true (snd scx) (n - 1) with
+                    | Some {core = Defn _} -> [Dx n @@ dw]
+                    | _ ->
+                        Errors.err ~at:dw
+                            "%S is not a definition and cannot be used in a \
+                            DEF clause."
+                            v;
+                        []
+                    end
                 | _ ->
                     Errors.warn ~at:dw
                         "Ignored unexpandable \
--- a/test/bugs/def_nondef_test.tla
+++ b/test/bugs/def_nondef_test.tla
@@ -1,0 +1,33 @@
+---- MODULE def_nondef_test ----
+
+(* Regression test for https://github.com/tlaplus/tlapm/issues/280
+
+   A DEF clause (`BY ... DEF X`, `USE DEF X`, or `HIDE DEF X`) that cites a
+   name which is in scope but is not a definition -- a CONSTANT, a VARIABLE, or
+   a NEW-declared constant/operator -- used to crash the whole run with
+   `Failure("Proof.Gen.set_defn/doit")` ("Oops, this seems to be a bug in
+   TLAPM"). It must instead be reported as a normal located error and the run
+   must not abort abnormally.
+*)
+
+CONSTANT C
+VARIABLE v
+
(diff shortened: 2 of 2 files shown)


lemmy:
In general, Errors.err prints to stderr but doesn't change the exit code, so citing a non-definition in a DEF clause still exits 0.  Given the input is an outright user error, should this be a hard failure (nonzero exit / status:failed) instead of a silently-continuing message?

lemmy (approved):
LGTM, but I hope others shed light on warn vs err vs fatal (https://github.com/tlaplus/tlapm/pull/282#issuecomment-4958970696).

muenchnerkindl:
LGTM, like @lemmy I do not really understand how the code uses Errors.err and Errors.fatal. But that's a more global issue that is not directly related to this PR.

PR #283, opened 2026-07-13
Add a deterministic Z3 `rlimit` budget via `SMTT("rN")`

### LINKED ISSUE #281: Add a deterministic Z3 `rlimit` budget for reproducible verification across machines and rerun
### Description

The `SMTT(n)` proof directive bounds the SMT backend by a **wall-clock** timeout of `n` seconds. Because wall-clock time depends on CPU speed, load, and parallelism, the same obligation can discharge within `n` seconds on one run and time out on another. The proof's pass/fail outcome then depends on the environment, so it cannot be shared, re-checked in CI, or reproduced on another host with confidence.

In one of my own experiments, **~95% of the timeouts came from Z3**, so making the Z3 budget deterministic would remove most of the non-determinism on its own. See also #89 and #53, which report symptoms of the same wall-clock timeouts.

### Proposal — deterministic budget via Z3 `rlimit`

Z3 exposes `rlimit`, a **resource counter**: it counts internal "resource units" consumed by the solver during search. For a fixed Z3 build and identical input, the resource count is deterministic and independent of CPU speed and load, so a bound expressed in `rlimit` yields the *same* pass/fail decision on every machine and every rerun.

The proposal is to let a proof express its SMT budget as an `rlimit`, so that a proof's outcome is a property of the proof and the Z3 build, and reproduces on any machine.

Concretely, when invoking Z3, pass the budget as `(set-option :rlimit N)` (equivalently `rlimit=N` on the command line).

### Design

New directive `SMTR(rlimit)` (and, analogously, `Z3R(rlimit)`), parallel to the existing `SMTT(seconds)` family. This keeps the two notions of budget explicit and lets a proof mix them.

### Scope and caveats

- **Z3-specific.** `rlimit` is a Z3 mechanism, so other backends (Zenon, Isabelle, CVC, veriT, …) keep their wall-clock timeout. This is a minor limitation in practice: as noted above, ~95% of the timeouts were Z3, so covering the Z3 path removes most of the non-determinism.
- **Fixed Z3 build.** `rlimit` values are reproducible for the Z3 tlapm ships; they may shift only if that Z3 is upgraded.

tangruize:
## Description

Closes #281.

Following @lemmy's suggestion on the issue, this reuses the existing `SMTT`/`Z3T` directives with a string argument instead of adding new `SMTR`/`Z3R` directives.

### What it does

`SMTT`/`Z3T` now accept a string budget of the form `"rN"` that bounds the underlying Z3 solver by a deterministic `rlimit` (resource-count) budget instead of a wall-clock timeout. For a fixed Z3 build the resource count — and therefore the pass/fail outcome — is the same on every machine and every rerun, so a proof's result no longer depends on CPU speed or load.

### Usage

```tla
THEOREM ... BY SMTT("r5")    \* deterministic Z3 budget, ~5 base units
THEOREM ... BY Z3T("r10")    \* same, via the Z3 directive
THEOREM ... BY SMTT("r0.5")  \* fractional budgets are allowed
```

`N` is a multiplier over a fixed base resource count: larger `N` means a larger budget. It is not wall-clock time — only the resource count is deterministic; how long it takes to consume the budget still varies by machine.

### Default / backward compatibility

Unchanged. The plain forms — `SMT`, `SMTT(n)`, `Z3`, `Z3T(n)` — set **no** `rlimit`, so Z3 uses its own default (`rlimit = 0`, i.e. unlimited) and the proof is bounded only by the wall-clock timeout (5 s by default, or `n` seconds). The deterministic budget is strictly opt-in via the `"rN"` string; without it, behaviour is exactly as before (and still non-deterministic across machines).

### Details

- `"rN"` is scaled to a raw Z3 count by `rlimit_base = 3_000_000` (mirroring [Verus's `RLIMIT_PER_SECOND`](https://github.com/verus-lang/verus/blob/457dca62be432460c816c03ece17582d28d39a91/source/rust_verify/src/verifier.rs#L48)) and clamped to Z3's unsigned-32-bit maximum, so a small, readable budget like `"r5"` is meaningful.
- The budget is threaded through the `Smt3`/`Z33` methods and emitted as `(set-option :rlimit N)` into the SMTLIB2 input.
- When an `rlimit` is given the wall-clock timeout is disabled so the deterministic budget alone bounds Z3, and `--stretch` does not scale it.
- Z3-specific: only the SMT/Z3 backends accept it. A malformed budget string, or an `rlimit` on a non-Z3 backend, is rejected with a normal located error.
- Fingerprints are unchanged: the `rlimit` is not recorded (a proved obligation stays proved regardless of the budget), so there is no fingerprint-format change.

### Tests

- `test/fast/basic/smt_rlimit_test.tla` — the proving path (`SMTT("rN")` / `Z3T("rN")` discharge obligations).
- `test/fast/basic/smt_rlimit_bad_test.tla` — a malformed budget is reported as a located error.


src/backend/prep.ml
```diff
@@ -1106,7 +1113,34 @@ let compute_meth def args usept =
...
+      Errors.err ~at
+        "Invalid SMT budget %S: expected a number of seconds (wall-clock \
+         timeout), or a deterministic Z3 rlimit budget of the form \"rN\" \
+         (e.g. \"r5\")."
+        s;
+      raise Exit
+    in
+    let n = String.length s in
+    if n >= 2 && (s.[0] = 'r' || s.[0] = 'R') then
+      match float_of_string_opt (String.sub s 1 (n - 1)) with
```
lemmy, line 1134: This underflows to `rlimit 0` for small resource limits like `r0.0000001` due to `int_of_float` truncating towards zero.

muenchnerkindl:
Thank you for this PR. My main question is how ``SMTT(...)`` behaves with respect to fingerprinting? It seems to me that time limits are fingerprinted because I recall having seen messages referring to previous attempts with different time bounds. The potential issue I see is that (when used from a standard IDE) the PM will ignore an attempt to prove a step that previously failed for a lower resource bound.

src/backend/prep.ml
```diff
@@ -1187,7 +1227,14 @@ let compute_meth def args usept =
   begin try read def args;
   with Exit -> failwith "error in prover specification"
   end;
-  match !prover with  (* FIXME should factor with Params.mk_meth *)
+  (* When a Z3 `rlimit` budget is given, the wall-clock timeout is disabled
+     (set to infinity) so that the deterministic budget alone bounds Z3. *)
+  let smt_budget default =
+    match !rlimit with
+    | Some _ -> (infinity, !rlimit)
```
lemmy, line 1237: Setting wall-clock timeout to infinity prevents tlapm from killing a rogue z3 where rlimit doesn't halt z3. Consider a large cap so rlimit still decides the verdict but a stuck z3 still gets killed.

src/backend/prep.ml
tangruize: Fixed in this commit: https://github.com/tlaplus/tlapm/pull/283/commits/684836e8de26d12a7d8dd725fdb0a8c4e402b857

tangruize:
> Thank you for this PR. My main question is how `SMTT(...)` behaves with respect to fingerprinting? It seems to me that time limits are fingerprinted because I recall having seen messages referring to previous attempts with different time bounds. The potential issue I see is that (when used from a standard IDE) the PM will ignore an attempt to prove a step that previously failed for a lower resource bound.

@muenchnerkindl Thanks! Fixed in this commit: https://github.com/tlaplus/tlapm/pull/283/commits/19c89cef0f6b5d64b2cdd8c6ebd3f7d5f90e92f2

ahelwer:
Sorry I posted on the wrong PR! Got mixed up. Anyway I had been dreaming of a feature like this for SMT solvers for a long time, had no idea it was actually implemented.

lemmy:
@tangruize Could you please address the DCO warning above?

muenchnerkindl:
@tangruize Could you please amend your commit to sign off (DCO)?

tangruize:
Thanks @lemmy @muenchnerkindl! I've addressed the DCO warning.

PR #284, opened 2026-07-22
Kill orphaned prover processes on Linux when tlapm dies

tangruize:
## Problem

tlapm launches every backend prover (z3, cvc4, zenon, …) as a child process and terminates it as part of its own cleanup. If tlapm dies without getting to run that cleanup — an uncatchable `SIGKILL`, or an OOM-kill under memory pressure — the prover it spawned is reparented to `init` and keeps running. Most solvers impose no limit on themselves, so the orphan goes on consuming CPU and memory until it happens to finish on its own.

## Fix

On Linux, run each prover through util-linux `setpriv --pdeathsig KILL`, which arms the kernel parent-death signal (`PR_SET_PDEATHSIG`): the kernel sends `SIGKILL` to the prover the moment its parent (tlapm) dies, including the `SIGKILL` and OOM cases where tlapm can no longer act itself.

The leading `exec` makes the prover replace the shell that launches it, so the prover becomes tlapm's direct child — the process the parent-death signal actually watches — instead of a grandchild under an intermediate `/bin/sh`. Without it that shell would stay alive and keep the prover running.

The prefix is added at the single point where every backend command string is assembled, so all provers are covered at once. On non-Linux, or on Linux where `setpriv` is missing or too old to support `--pdeathsig`, the prefix is empty and behaviour is unchanged.

### DIFF
--- a/src/params.ml
+++ b/src/params.ml
@@ -102,6 +102,34 @@ type exec = executable ref
 let path_prefix =
   sprintf "PATH='%s';" Paths.backend_path_string
 
+(* Prefix that makes a prover process die together with tlapm.
+
+   On Linux we run the prover through util-linux `setpriv --pdeathsig KILL`,
+   which arms the kernel's parent-death signal (PR_SET_PDEATHSIG): the kernel
+   sends SIGKILL to the prover as soon as its parent (tlapm) dies, including
+   when tlapm is itself SIGKILLed or OOM-killed and can no longer run any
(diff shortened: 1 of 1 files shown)


lemmy (approved):
LGTM

lemmy:
This PR is Linux-specific, which is completely reasonable. That said, I’m seeing a similar issue on macOS (when using AI to drive TLAPS). It would be great to address the same problem on macOS and Windows in a future follow-up.

Issue #287, opened 2026-07-29
tlapm exits 0 when proof obligations fail, so CI reports success on a broken proof


vasilisnasopoulos:
## Summary

When obligations fail, `tlapm` reports it clearly — in the prose summary and in
the machine-readable `@!!status:failed` markers — but still exits with status
`0`. Any CI system that treats the exit code as its success signal will go green
over a proof that did not close.

This is not a soundness issue: tlapm never claims a failed obligation was
proved. The information is present in every channel *except* the one automation
reads.

Reproduced on both the current stable release and today's rolling pre-release.

## Minimal reproducer

```tla
---- MODULE ExitCodeRepro ----
EXTENDS Naturals

THEOREM Provable == 1 = 1
  OBVIOUS

THEOREM Unprovable == 1 = 2
  OBVIOUS

====
```

```sh
tlapm --toolbox 0 0 ExitCodeRepro.tla > out.log 2>&1
echo "exit: $?"
```

| Version | Output | Exit status |
|---|---|---|
| 1.5.0 (release `202210041448`) | `[ERROR]: 1/2 obligations failed.` | **0** |
| 1.6.0-pre (`096df83`) | `[ERROR]: 1/2 obligations failed.` | **0** |

Both emit three `@!!status:failed` markers for the single failed obligation
(presumably one per backend attempt), so the failure is visible in the
structured output as well.

Full CI run of the above, both versions side by side:
<https://github.com/vasilisnasopoulos-stack/vortex-dse-cslot-proofs/actions/runs/30478555237>

That workflow run is itself reported as **success** by GitHub Actions.

## How this played out in practice

Before reducing it to the case above, I hit it for real. This run is green:

<https://github.com/vasilisnasopoulos-stack/vortex-dse-cslot-proofs/actions/runs/28464335410>

- GitHub Actions conclusion: **success**
- In the log: `[ERROR]: 1/99 obligations failed.`

The step was `tlapm --toolbox 0 0 <file>.tla` with no output parsing, so it
passed on exit status alone. A public repository advertised a green proof badge
over a broken proof for four days, until I happened to read the log by hand.

`1.5.0` is still the latest stable release, and the usual CI recipe installs it
directly:

```sh
wget -q https://github.com/tlaplus/tlapm/releases/download/202210041448/tlaps-1.5.0-x86_64-linux-gnu-inst.bin
```

So this is the default experience for anyone wiring TLAPS into CI today.

## Relation to #276

[...]

PR #288, opened 2026-07-30
doc: align Download/Binaries pages with README

vasilisnasopoulos:
Follows up on #287, where you asked for a PR aligning `Download/Binaries/*` with `README.md`.

`README.md` directs readers to the 1.6.0 rolling pre-release for the current version, and to the releases page for past versioned releases. The Binaries pages still presented the 1.5.0 installer (release `202210041448`, October 2022) as "The package", with no indication that anything newer exists.

**Changes**

- `Binaries/Linux.html` and `Binaries/MacOS.html` now lead with the rolling pre-release tarball: download link, unpack, `PATH`, and the location of `TLAPS.tla` for the Toolbox / VS Code library path. The existing 1.5.0 installer instructions are kept unchanged, under a dated heading.
- `Binaries/MacOS.html` additionally notes that the pre-release currently carries an Apple-silicon build only, so Intel users should build from source or use 1.5.0. The page previously offered only `tlaps-1.5.0-i386-darwin-inst.bin`.
- `Binaries.html`: "Download the TLAPS installer" → "Download TLAPS", since the current distribution is a tarball rather than an installer.
- `Binaries/Windows.html` is unchanged: it defers to the Linux page, so it picks this up automatically.

**Verified against the published artifact** rather than assumed — extracting `tlapm-1.6.0-pre-x86_64-linux-gnu.tar.gz` gives a relocatable tree with `tlapm/bin/tlapm` and `tlapm/lib/tlapm/stdlib/TLAPS.tla`, and `tlapm --config` resolves its paths relative to wherever it was unpacked. The tarball contains no `examples/` directory, so the old "copy the example files" step was not carried over to the new section.

Happy to adjust the wording, or to split this differently if you would rather the pages point at the releases page generically instead of naming `1.6.0-pre`.

### DIFF
--- a/doc/web/content/Download/Binaries.html
+++ b/doc/web/content/Download/Binaries.html
@@ -19,8 +19,8 @@ are missing the navigation menus.</b></p><hr/></noscript>
 <!-- DO NOT EDIT ABOVE THIS LINE, DO NOT REMOVE THIS LINE -->
 
 <div class="section">
-  <h2>Download the TLAPS installer</h2>
-  <p class="first">The TLAPS binary installer is available for the
+  <h2>Download TLAPS</h2>
+  <p class="first">Binary builds are available for the
   following systems:
     <ul style="list-style-type:none;list-style:none">
       <li>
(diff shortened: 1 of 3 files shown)


lemmy (approved):
LGTM

Issue #181, opened 2024-11-27
Reduce size of TLAPM release bundle


ahelwer:
Currently, TLAPM releases are enormous; they take up 473 MB zipped, and 1.1 GB unzipped. This seems far, far larger than should be necessary. Unzipping a TLAPM release, we have the following notably-large components:
```
/
├── (57  MB) bin/
│   ├── (17  MB) tlapm
│   ├── (37  MB) tlapm_lsp
│   └── (3.6 MB) translate
└── (1.1 GB) lib/tlapm/backends/
    ├── (36  MB) bin/
    │   ├── (1.5 MB) ls4
    │   ├── (3.6 MB) ptl_to_trp
    │   ├── (26  MB) z3
    │   └── (5.1 MB) zenon
    └── (1.1 GB) Isabelle/
        ├── (33  MB) heaps/polyml-5.9.1_x86_64_32-linux/
        │   ├── (23  MB) Pure
        │   └── (9.1 MB) TLA+
        ├── (71  MB) src/
        ├── (9.6 MB) lib/classes/
        │   └── (9.6 MB) isabelle.jar
        └── (903 MB) contrib/
            ├── (355 MB) jdk-21.0.3
            ├── (84  MB) stack-2.15.5
            ├── (61  MB) mlton-20210117-3
            ├── (56  MB) jedit-20240425
            ├── (48  MB) polyml-5.9.1
            ├── (36  MB) naproche-20240519
            ├── (27  MB) e-3.0.03-1
            ├── (23  MB) vampire-4.8
            ├── (21  MB) cvc4-1.8
            ├── (19  MB) pdfjs-2.14.305
            ├── (17  MB) jfreechart-1.5.3
            ├── (15  MB) zipperposition-2.1-1
            ├── (14  MB) z3-4.4.0pre-4
            ├── (14  MB) sqlite-3.45.2.0
            ├── (11  MB) rsync-3.2.7-1
            ├── (6.7 MB) zstd-jni-1.5.5-4
            ├── (6.0 MB) verit-2021-06.2-rmx-1
            └── (5.9 MB) opam-2.0.7
```
There are a few smaller candidates for removal:
- `tlapm_lsp`; does it need to be there it should it be a separate OPAM package?
- The entirety of the `Isabelle/src` directory

[...]

PR #292, opened 2026-09-08
Avoid including full Isabelle to the TLAPM release.

kape1395:
Only the `polyml` and the pre-built heaps along with several config files are needed to invoke Isabelle TLA+ theories.
This also allowed to simplify the build/release by removing the post-install script.
The invocation is now similar to that of `--fast-isabelle`, which is now the only way to run it.

The release archive is now **47MB** instead of **843MB** and extracted it takes 145MB instead of 1518MB.

Acknowledgements: AI was used while developing this PR, but I'm responsible for the changes.


kape1395:
Closes https://github.com/tlaplus/tlapm/issues/181

kape1395:
@muenchnerkindl , maybe you could take a look at this? This is mostly the build-related changes.
Probably the only places worth looking:
  - the Isabelle invocation (https://github.com/tlaplus/tlapm/pull/292/changes#diff-622154c57af633b5972075b1e1277b7c04a5241427dcb630e35d92f459430425R159), which is similar to what was done for cygwin / --fast-isabelle,
  - how the Isabelle options are built into a heap `Options` here: https://github.com/tlaplus/tlapm/pull/292/changes#diff-4a2938c80d0975ccb96c7a1e65aa89e7d450be48d0798c5640b034960ae898f6R100

